PR #13, opened 2019-05-24
Os app killswitch conan qt

### LINKED ISSUE #11: Use pre-built Qt on all platforms

jmarrec:
Fix #11

I merged all active branches: `OS_app_killswitch`, `OS_App_killswitch_packaging `and `Move_ruby_modeleditor`.

Then I enabled conan-qt. It builds on my Ubuntu 16.04 machine, but I expect there will be problems.

Packaging still needs some deep attention, there are a lot of things that are still in openstudio (core) that shouldn't. Allow will need to figure out how to have this work on fresh machines for Qt deps.

**Note: I made the git submodule track my jmarrec:OpenStudio/OS_App_killswitch_Qt for now**

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,5 +1,5 @@
 [submodule "openstudio"]
 	path = openstudio
-	url = git@github.com:jmarrec/OpenStudio.git
+	url = git@github.com:NREL/OpenStudio.git
 	branch = OS_App_killswitch_Qt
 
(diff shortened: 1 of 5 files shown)


macumber:
@jmarrec this is dead now right?  We should just close?

jmarrec:
Yeah let's close. I don't think we want to move to conan Qt (yet, at least). We'll need to remember some work was done already if we do (cf https://github.com/jmarrec/conan-openstudio-qt)

PR #49, opened 2019-11-28
Test code reuse

jmarrec:
cf https://github.com/NREL/cbci_jenkins_libs/pull/49


macumber:
This branch is dead right?  We should just close?

jmarrec:
This was specifically to test some changes to Jenkins CI (I had created a groovy module to use the same code whether on mac, linux or windows).

PR #93, opened 2020-04-06
Fix #92 - Restore the style of the OSApp on Windows

### LINKED ISSUE #92: Application style is old fashioned on Windows: regression to "XP" style.
Issue overview
--------------

The "style" / look and feel of the GUI App on Windows has regressed.

## Current Behavior

This is what the rc1 installer does on Windows right now
image (2)

## Expected Behavior

This is what 2.9.1 looked like on Windows:

image

This is what Linux keeps looking like:

image (1)

## Possible Solution

Likely a change when we bumped Qt from 5.6 to Qt 5.11

### Environment

 - Platform (Operating system, version): Windows
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): 1.0.0-rc1: 12d1189eabb1156ad3a13b6769572f391e993257

jmarrec:
Fix #92 - In Qt 5.11, unlike the 5.6 we used before, we need to explicitly ship the qwindowsvistastyle.dll on Windows to avoid having an "XP-like" UI.


jmarrec:
Tested locally on windows @tijcolem , it works fine. Testing showed I had to make_directory styles/ first, now it builds, and after launch the style is restored.

PR #182, opened 2020-06-18
Fix linking error in openstudio_lib_tests

jmarrec:
Getting a link error in openstudio_lib_tests:

```
[ 95%][544/570] Linking CXX executable Products/openstudio_lib_tests-1.0.0
FAILED: Products/openstudio_lib_tests-1.0.0 
[...]

PR #185, opened 2020-06-29
Fix #184 - Support WaterCooled VRFs in OSApp

### LINKED ISSUE #184: Support WaterCooled VRF in OS App
Enhancement Request
--------------

https://github.com/NREL/OpenStudio/pull/3613 added WaterCooled `OS_AirConditioner_VariableRefrigerantFlow` to the SDK. OSApp doesn't properly support it yet.

## Possible Implementation

* OpenStudioPolicy.xml needs to be updated to hide the Condenser Inlet / Outlet Nodes.
* Icon isn't registered, (seems like this commit didn't make it in : https://github.com/NREL/OpenStudio/commit/335ab90e3cca3d0939a68a5cd18a08851db6fbb5)
* MainRightColumnController::configureForHVACSystemsSubTab for "My Model" needs to add the VRF System.

jmarrec:
Fix #184  - Support WaterCooled VRFs in OSApp

### DIFF
--- a/src/openstudio_lib/IconLibrary.cpp
+++ b/src/openstudio_lib/IconLibrary.cpp
@@ -75,6 +75,7 @@ IconLibrary::IconLibrary()
   // Default icon to avoid a crash
   m_icons[openstudio::IddObjectType(openstudio::IddObjectType::Catchall).value()] = new QPixmap(":images/missing_icon.png");
 
+  m_icons[openstudio::IddObjectType(openstudio::IddObjectType::OS_AirConditioner_VariableRefrigerantFlow).value()] = new QPixmap(":images/vrf_outdoor.png");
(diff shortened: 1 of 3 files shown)


src/openstudio_lib/CMakeLists.txt
```diff
@@ -716,7 +716,7 @@ add_library(${target_name}
   ${${target_name}_qrcs}
 )
 
-set(depends
+set(${target_name}_depends
```
macumber, line 719: This looks like a decent change, why was it causing an issue?  Is there another variable named `depends`?

macumber: Maybe this one `osapp\src\openstudio_app\CMakeLists.txt`?

jmarrec: This is from https://github.com/openstudiocoalition/OpenStudioApplication/pull/182

PR #187, opened 2020-07-09
Add CI via GitHub Actions - cross platform, github-hosted.

jmarrec:
Try a build with the following characteristics:

* Cross platform (mac, windows, linux)
* Caches the OpenStudio SDK (later we would rely on CMake-foo to deal with it instead of actions with jq)


macumber:
@jmarrec FYI https://github.com/openstudiocoalition/OpenStudioApplication/tree/github_actions

jmarrec:
> @jmarrec FYI https://github.com/openstudiocoalition/OpenStudioApplication/tree/github_actions

Yeah, I know, I named my branch 'github_actions' too and my git client is the one that told me you had done something at the same time:

image

jmarrec:
Almost there after a lot of effort... As part of this PR dependencies were bumped to rely on conan-center-index only and to match conan-openstudio-ruby and NREL/OpenStudio open PRs.

.github/workflows/app_build.yml
```diff
@@ -0,0 +1,304 @@
+name: C/C++ CI for OpenStudioApplication
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
macumber, line 7: Suggest running CI on pull requests to master or develop and on push to master

macumber: These might be different, we might want to make tags on push to master

.github/workflows/app_build.yml
```diff
@@ -0,0 +1,304 @@
...
+
+env:
+  OS_SDK_VERSION: 3.0.1
```
macumber, line 10: `OS_SDK_VERSION` should be pulled from a file in source, something like `cat OSVersion.txt`, this is a bit too buried

macumber: Or pull it from `CMakeLists.txt` like you did for app version.

.github/workflows/app_build.yml
```diff
@@ -0,0 +1,304 @@
...
+      # To avoid downloading the SDK all the time, we try to cache it
+      id: cacheossdk
+      uses: actions/cache@v2
+      with:
+        path: OpenStudio-${{ env.OS_SDK_VERSION }}
+        key: OpenStudio-SDK-${{ runner.os }}-${{ env.OS_SDK_VERSION }}
+
+    - name: Download the OpenStudio installer
+      # If the SDK wasn't found in the cache
+      if: steps.cacheossdk.outputs.cache-hit != 'true'
```
macumber, line 130: Wow, this is cool!  I had no idea this was an option.

.github/workflows/app_build.yml
```diff
@@ -0,0 +1,304 @@
...
+      # Use a bash shell so we can use the same syntax for environment variable
+      # access regardless of the host operating system
+      # Note: wget not available on windows, even in bash, so use curl
+      shell: bash
+      run: |
+        curl --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' https://api.github.com/repos/NREL/OpenStudio/releases/tags/v$OS_SDK_VERSION > os_tag.json
```
macumber, line 136: We probably need a way to point this at a branch or even a fork of the SDK

.github/workflows/app_build.yml
```diff
@@ -0,0 +1,304 @@
...
+          echo "Conan cache not found"
+
+    - name: Did restoring the conan-cache work? Yes
+      # If the SDK wasn't found in the cache
+      if: steps.cacheconan.outputs.cache-hit == 'true'
+      run: |
+          ls ~/.conan/
+          ls ~/.conan/data/
+
+    - name: Cache entire build directory
```
macumber, line 188: Caching the entire build directory might not be worth it, we tried this on local CI and found it took longer to transfer the build files than it took to rebuild.

src/openstudio_lib/library/OpenStudioPolicy.xml
```diff
@@ -1236,7 +1230,7 @@
     <rule IddField="Nominal Floor to Ceiling Height" Access="LOCKED"/>
   </POLICY>
   <POLICY IddObjectType="OS_DefaultConstructionSet">
-    <rule IddField="Name" Access="LOCKED"/>
+    <!--<rule IddField="Name" Access="LOCKED"/>-->
```
macumber, line 1233: What's this about?

.github/workflows/app_build.yml
jmarrec: A fork would proabably be too burdensome, because we'd have to actually get github actions to **build** the SDK before building the app. I doubt we want that.

A PR to OpenStudio is _doable_, since NREL's CI (jenkins) will build and upload installers that we could use, but that's going to be a lot of work to get _just_ right:

* you need an easy way to be able to specify which SDK PR to go reference... perhaps in the OpenStudioApplication PR message. I've thought about it for Jenkins a few months back - I would create a PR template so it could be reproducible
* then locate the right installer from that (and that is harder)

src/openstudio_lib/library/OpenStudioPolicy.xml
jmarrec: From another PR: https://github.com/openstudiocoalition/OpenStudioApplication/pull/107/files#diff-3f25226a4f407224991996e30426d766R1233

.github/workflows/app_build.yml
jmarrec: This is something we can adjust easily after the fact (by simply commenting it out), after we determine whether:

1. Caching actually works
2. Caching is actually beneficial

jmarrec: I can see the cache being actually restored in under 3 secs, but I don't think the **build** step is actually faster because of it. Not sure why.

.github/workflows/app_build.yml
jmarrec: I pulled it from FindOpenStudioSDK.cmake...
At which point you start to wonder whether getting CMake to download the tar.gz instead of using the bash script I wrote would make more sense...

.github/workflows/app_build.yml
jmarrec: Done, we can adjust later.

jmarrec:
Saving the combined Static analyzer workflow for reference, but deciced to separate cppcheck and clang-format:

```
name: Static Analyzers

on:
  pull_request:
    branches: [ master, develop ]

jobs:
  cppcheck:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: manual cppcheck
      shell: bash
      run: |
          sudo snap install cppcheck
          cppcheck --suppress=missingInclude --enable=all --inline-suppr --inconclusive --template='[{file}:{line}]:({severity}),[{id}],{message}' --force -q -j $(nproc) ./src  2>&1 | tee cppcheck.txt

    - name: Parse and colorize cppcheck
      shell: bash
      run: python ./ci/colorize_cppcheck_results.py

    - name: Upload cppcheck results as artifact
      if: ${{ always() }}
      uses: actions/upload-artifact@v2
      with:
          name: OpenStudioApplication-${{ github.sha }}-cppcheck_results.txt
          path: cppcheck.txt

  clang-format:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v2

    - name: Run clang-format against C++ files touched by the PR
      shell: bash
      run: ./ci/clang-format.sh remotes/origin/$GITHUB_HEAD_REF remotes/origin/$GITHUB_BASE_REF

    - name: Upload clang-format patch as artifact
      if: ${{ failure() }}
      uses: actions/upload-artifact@v2
      with:
          name: OpenStudioApplication-${{ github.sha }}-clang_format.patch
          path: clang_format.patch

```

jmarrec:
I added cppcheck with a custom parser/colorizer that I wrote in python.
And I added a clang-format file, and a clang-format github actions myself too.

jmarrec:
@macumber I'm merging. I have no time left to donate on this and it appears to work fine. We can tweak it later.

PR #196, opened 2020-07-22
Run clangformat accross all code base

jmarrec:
and also ignore src/qtwinmigrate from the cppcheck

```bash
all_files=$(find src/ -name '*' | /bin/grep '.*\.\(cpp\|c\|cxx\|cxx.in\|hpp\|h\|hxx\|hxx.in\)$')
echo $all_files | xargs clang-format -style=file -i -fallback-style=none
```


jmarrec:
I think I'd prefer if `public` or `private` would have a different indent than the functions in that scope

src/.clang-format
```diff
@@ -1,6 +1,6 @@
 ---
 Language:        Cpp
-AccessModifierOffset: -4
```
macumber, line 3: +1

jmarrec:
I added a .git-blme-ignore-revs file. Make sure locally you run `git config blame.ignoreRevsFile .git-blame-ignore-revs` to take it into account. It doesn"t work on github, but locally it will. Use the `[chore]` tag in the commit message when you do stuff (eg: update copyrights, or run automated tool) that should be filtered out.

Issue #177, opened 2020-05-14
OpenStudioApp Windows crashed during save attempt


tijcolem:
A user reported that when using OpenStudioApplication v1.0.0 the application crashes when trying to save the osm. See video at 5:15: https://www.youtube.com/watch?v=swX0ewW325g&t=542s

This very well could be a legitimate permissions problem on Windows when boost is trying to create a directory, but this should be a handled exception vs crash the app. 

This is using Windows: https://github.com/NREL/OpenStudioApplication/releases/download/v1.0.0/OpenStudioApplication-1.0.0.4f5416cfc7-Windows.exe

jmarrec:
Duplicate of #59 that has more info

rwadhwa910:
@tijcolem  I am having the same issue on the newer version of Open studio 3.0.1. I have attached the screenshot.

@DavidGoldwasser

rwadhwa910:
Apart from the crash in the app, When I open the same model again in open studio, I am not able to edit the model in editor tab. The only option I see is of creating a new model.

DavidGoldwasser:
Just to update @rwadhwa910 its comment, the floorplan.json and I believe other files (measures, weather?) have been deleted from the project.

rwadhwa910:
@DavidGoldwasser Weather file was deleted, I am not sure whether measure file was deleted, If that happens again I will note it and update it on this page.

DavidGoldwasser:
Just wanted to share another video demonstrating this by TheEnergyTutor on UH. He appears to not have the directory open, but still gets the crash related to file permission. I could not recreate this on my windows machine following same steps that user did. I had hoped that maybe I could recreate it by deleting a measure in the GUI like he did, but it wasn't reproducible for me.
https://unmethours.com/question/46008/dview-and-severe-crash-help/

macumber:
This appears to still be valid:

https://unmethours.com/question/45931/is-anyone-have-problems-with-floorspace-js-losing-floor-plans/
https://unmethours.com/question/45913/another-unhandled-exception/

nklammer:
I just wanted to add that the trigger for this problem seems to also be having the GitHub Desktop app open during simulation via the OpenStudio App 1.0.1.

macumber:
Was your OSM model directory under version control?  If so GitHub Desktop app was probably locking the files or something like that.  Have you seen this in v1.1.0 or higher?

PR #197, opened 2020-07-26
Fixes crash on save when model directory is open in Explorer on Windows

macumber:
Fixes https://github.com/openstudiocoalition/OpenStudioApplication/issues/177

Depends on https://github.com/NREL/OpenStudio/pull/4026

We should decide if we want to fork OpenStudio to get this change in the app earlier, if we want to wait for NREL to make a release with it, or something else

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,8 @@ build_profile/
 build-profile/
 core-build/
 super-build/
+debug/
+release/
 /super-build-shared/
 /super-build-static/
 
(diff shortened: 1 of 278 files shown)


macumber:
@jmarrec I branched this from your fix clang format commit which I thought was already on develop

macumber:
Closed in favor of https://github.com/openstudiocoalition/OpenStudioApplication/pull/198

macumber:
Re-opened because other PR has clang format issues

jmarrec:
Ran clang format on #198, so use that instead.

PR #208, opened 2020-08-08
Allow pulling openstudio archives from either the official release page or the nightly develop builds. Update workflow to use that.

macumber:

PR #221, opened 2020-08-24
North angle

### LINKED ISSUE #214: North Axis not synchronized between OSM and Floorspace JSON
App seems to be using old `config["initialNorthAxis"]` key to store north angle:

https://github.com/openstudiocoalition/OpenStudioApplication/blame/develop/src/openstudio_lib/GeometryEditorView.cpp#L273

Correct key is `project["north_axis"]`:

https://github.com/NREL/OpenStudio/blob/develop/src/utilities/geometry/FloorplanJS.cpp#L1508
https://github.com/NREL/floorspace.js/blame/develop/schema/geometry_schema.json#L122

macumber:
Fixes #214

Depends on https://github.com/NREL/OpenStudio/issues/4054

### DIFF
--- a/src/openstudio_lib/GeometryEditorView.cpp
+++ b/src/openstudio_lib/GeometryEditorView.cpp
@@ -191,7 +191,7 @@ FloorspaceEditor::FloorspaceEditor(const openstudio::path& floorplanPath, bool i
   if (site) {
     m_originalSiteName = site->nameString();
   }
-  
+
   if (exists(m_floorplanPath)) {
 
     openstudio::filesystem::ifstream ifs(m_floorplanPath);
@@ -212,7 +212,7 @@ FloorspaceEditor::FloorspaceEditor(const openstudio::path& floorplanPath, bool i
       if (building) {
         m_floorplan->setNorthAxis(-building->northAxis());
       }
-  
+
       // do not update floorspace lat/lon from model, location in model probably comes from an epw file
 
       QString errorsAndWarnings;
@@ -521,8 +521,7 @@ void FloorspaceEditor::translateExport() {
     m_exportModelHandleMapping = rt.handleMapping();
 
     // manually add mappings between current model and new model for Site, Facility, and Building objects
-    m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Site>().handle()] = 
-      m_exportModel.getUniqueModelObject<model::Site>().handle();
+    m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Site>().handle()] = m_exportModel.getUniqueModelObject<model::Site>().handle();
     m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Facility>().handle()] =
       m_exportModel.getUniqueModelObject<model::Facility>().handle();
     m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Building>().handle()] =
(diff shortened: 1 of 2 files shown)


jmarrec:
Obviously since CI uses OS 3.0.1 it will never build it correctly.

src/openstudio_lib/GeometryEditorView.cpp
```diff
@@ -196,6 +208,13 @@ FloorspaceEditor::FloorspaceEditor(const openstudio::path& floorplanPath, bool i
       model::FloorplanJSForwardTranslator ft;
       m_floorplan = ft.updateFloorplanJS(*m_floorplan, m_model, true);
 
+      // update north axis from model, floorspace js northAxis is opposite of EnergyPlus's
+      if (building) {
+        m_floorplan->setNorthAxis(-building->northAxis());
+      }
```
jmarrec, line 214: Hum...

src/openstudio_lib/GeometryEditorView.cpp
```diff
@@ -487,6 +519,14 @@ void FloorspaceEditor::translateExport() {
   if (model) {
     m_exportModel = *model;
     m_exportModelHandleMapping = rt.handleMapping();
+
+    // manually add mappings between current model and new model for Site, Facility, and Building objects
+    m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Site>().handle()] = m_exportModel.getUniqueModelObject<model::Site>().handle();
+    m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Facility>().handle()] =
+      m_exportModel.getUniqueModelObject<model::Facility>().handle();
+    m_exportModelHandleMapping[m_model.getUniqueModelObject<model::Building>().handle()] =
+      m_exportModel.getUniqueModelObject<model::Building>().handle();
```
jmarrec, line 528: Why is this needed? I'm not doubting it isn't, just curious.

jmarrec (approved):
As long as it builds and tests fine, I'm ok with it.

src/openstudio_lib/GeometryEditorView.cpp
macumber: This one is pretty weird.  Basically we don't have site, facility, or building objects in the FloorspaceJS file so we don't have anywhere to store the corresponding OpenStudio Model handles.

src/openstudio_lib/GeometryEditorView.cpp
macumber: What was the concern, the difference in north angle convention?

macumber:
Thanks for the review @jmarrec, it does build provided we hook it up with the current OS SDK package.  NREL merged https://github.com/NREL/OpenStudio/issues/4054 so we should be able to update to use the packages on that PR. I'm not 100% sure how to do that since my work on https://github.com/openstudiocoalition/OpenStudioApplication/pull/208 was incorrect.

src/openstudio_lib/GeometryEditorView.cpp
macumber: I did leave some comments in the OS SDK: https://github.com/macumber/OpenStudio/blob/ac01edda9d24618b9c552f623e0a5c92bdd5de86/src/model/ThreeJSReverseTranslator.hpp#L67

macumber:
Blocked by https://github.com/openstudiocoalition/OpenStudioApplication/issues/225

src/openstudio_lib/GeometryEditorView.cpp
jmarrec: Yeah.

jmarrec:
@macumber After #208, I rebased onto develop (which pulls OpenStudio 3.1.0-alpha) and force pushed, CI should now build it correctly.

jmarrec:
@macumber this seems fine and CI is fine with it too. I wonder if we couldn't benefit from tests though?

macumber:
Very cool, thanks @jmarrec!

Issue #239, opened 2020-09-19
Shading Control Enhancements


macumber:
Enhancement Request
--------------

## Possible Implementation
https://docs.google.com/presentation/d/1py1qnkI3TRodwc2oqypxDHh4oX0gzcF7dMcSaL5u2ic/edit?usp=sharing

PR #240, opened 2020-09-19
Update to 3.1.0 SDK - punt on new shading controls UI

macumber:
I feel bad since you put a lot of effort into moving the shading controls UI @jmarrec 

However, I feel that shading and daylighting controls in EnergyPlus are fundamentally broken.  I think we should push to correct that implementation at the EnergyPlus and/or OpenStudio level before trying to make a new UI.  

If we do make a new UI, I think that would be a larger new feature that would need support from the coalition or others.  I created https://github.com/openstudiocoalition/OpenStudioApplication/issues/239 and linked to a Google doc to play with ideas.

My suggestion is that we get our proposal on refactoring these controls ready, then bring to DOE/NREL and see if they can work with us on refactoring at the EnergyPlus and/or OpenStudio level


src/openstudio_lib/SpacesSubsurfacesGridView.cpp
```diff
@@ -431,8 +433,14 @@ void SpacesSubsurfacesGridController::addColumns(const QString& category, std::v
       } else if (field == SHADINGSURFACENAME) {
 
       } else if (field == SHADINGCONTROLNAME) {
-        addDropZoneColumn(Heading(QString(SHADINGCONTROLNAME)), CastNullAdapter<model::SubSurface>(&model::SubSurface::shadingControl),
-                          CastNullAdapter<model::SubSurface>(&model::SubSurface::setShadingControl),
+
+        // temporary workaround, see Shading Control Enhancements #239
+        std::function<bool(model::SubSurface*, const model::ShadingControl&)> setter(
+          [](model::SubSurface* t_surface, const model::ShadingControl& t_arg) {
+            return const_cast<model::ShadingControl&>(t_arg).addSubSurface(*t_surface);
```
macumber, line 440: `const_cast` is ugly but ok for now

PR #244, opened 2020-10-07
Update standards

macumber:
@jmarrec i am just trying to hack this to get it to run right now

PR #246, opened 2020-10-07
Update export openstudio-standards action until it works

jmarrec:
Ok, now that it's on develop, I can do stuff here. And it's running fine finally:

image

https://github.com/openstudiocoalition/OpenStudioApplication/runs/1221042646?check_suite_focus=true

### DIFF
--- a/.github/workflows/export_standards_data.yml
+++ b/.github/workflows/export_standards_data.yml
@@ -146,6 +146,7 @@ jobs:
                 exit 1
               fi
 
+              echo -e "${Yellow}Trying with latest release: $installer_link${Color_Off}"
               if ! wget -q "$installer_link"; then
                 echo -e "${Red}Not found at $installer_link${Color_Off}"
                 echo -e "${Red}I've exhausted all options here. Sorry${Color_Off}"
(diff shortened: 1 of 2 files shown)


.github/workflows/export_standards_data.yml
```diff
@@ -42,56 +45,81 @@ jobs:
 
         sdk_version = ''
         if m_major:
-            sdk_version += m_major.groups()[0]
-            print("::set-env name=OS_SDK_VERSION_MAJOR::{}".format(m_major.groups()[0]))
+            OS_SDK_VERSION_MAJOR = m_major.groups()[0]
```
macumber, line 48: I hesitate to say stuff like this because I don't want you to go off and nerd-snipe yourself into doing it, but downloading the OS SDK seems like it would be a good thing to factor into a separate action

macumber (approved):
Looks good, now we just need to edit the script to not crash on the DEER export? Or remove DEER?

.github/workflows/export_standards_data.yml
jmarrec: Yeah, we'll deal with that another day.

jmarrec:
> Looks good, now we just need to edit the script to not crash on the DEER export? Or remove DEER?

Moving that discussion to https://github.com/openstudiocoalition/OpenStudioApplication/pull/244. The github actions itself works.

PR #247, opened 2020-10-07
Fix #203 - Disable dark mode on mac

### LINKED ISSUE #203: [Mac] UI is unreadable with white font on grey background
Issue overview
--------------

Reported by @ljbrackney : 

>Is it just me, or does the 1.0 App use white on light gray text for pick lists on Mac?  It’s nearly unreadable.
> I thought maybe it was a light/dark mode thing, but it looks crummy no matter which I select.

## Current Behavior

image

## Expected Behavior

it should be legible!

## Steps to Reproduce

1.
2.
3.
4.

### Environment

 - Platform (Operating system, version): Mac
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): openstudioapplication 1.0.0

jmarrec:
Fix #203 - Disable dark mode on mac according to https://developer.apple.com/documentation/appkit/nsappearancecustomization/choosing_a_specific_appearance_for_your_macos_app#2993819

> The system automatically opts in any app linked against the macOS 10.14 or later SDK to both light and dark appearances. If you need extra time to work on your app's Dark Mode support, you can temporarily opt out by including the NSRequiresAquaSystemAppearance key (with a value of true) in your app’s Info.plist file. Setting this key to true causes the system to ignore the user's preference and always apply a light appearance to your app.


macumber:
Just tried it out and this works great

macumber (approved):

PR #256, opened 2020-10-17
Update to 3.1.0 release with patch for RemoteBCL bug

macumber:


macumber:
Built using packages from https://github.com/NREL/OpenStudio/pull/4121

jmarrec:
thanks for updating @macumber

macumber:
@jmarrec after CI finishes, I think we can merge this to develop, then master, then tag the release.  Release notes are not included in the package so we can just add them later to the GitHub release

PR #260, opened 2020-10-24
Replace deprecated set-env with $GITHUB_ENV

macumber:


macumber:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

.github/workflows/app_build.yml
```diff
@@ -138,23 +139,23 @@ jobs:
         sdk_version = ''
         if m_major:
             sdk_version += m_major.groups()[0]
-            print("::set-env name=OS_SDK_VERSION_MAJOR::{}".format(m_major.groups()[0]))
+            os.system(f"echo OS_SDK_VERSION_MAJOR={m_major.groups()[0]} >> $GITHUB_ENV")
```
jmarrec: That works probably. I've used it differently in python

https://github.com/openstudiocoalition/OpenStudioApplication/blob/e95d68c8d76868a736e1e3b5d3dfd804f8872ed8/.github/workflows/export_standards_data.yml#L50-L51

jmarrec: I'll bring that over from export_standards_data.yml which is more up to date on that front. We should put that code into its own reusable library file or something at one point, but no rush.

jmarrec (approved):
Thanks for taking care of that, it was on my todo

jmarrec:
```
Warning: The `add-path` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

PR #277, opened 2020-11-22
Fix loading openstudio_modeleditor.rb

macumber:
Upgrading to Qt 5.15.0 broke the way that we were loading openstudio_modeleditor.rb on Mac.  Qt apps are now relocatable and seemed to require that we install Qt libraries as a framework rather than just shared libraries.

Fixes https://github.com/openstudiocoalition/openstudio-sketchup-plugin/issues/67

Change to Qt https://www.qt.io/blog/qt-is-relocatable


.github/workflows/app_build.yml
```diff
@@ -27,7 +27,7 @@ env:
   CPACK_SOURCE_TXZ: OFF
   CPACK_SOURCE_TZ: OFF
   CPACK_SOURCE_ZIP: OFF
-  QT_VERSION: 5.15.0
+  QT_VERSION: 5.15.2
```
macumber, line 30:
```suggestion
  QT_VERSION: 5.15.0
```

macumber:
@jmarrec the Mac CI is having trouble with termcap, it looks like you pushed a fix https://github.com/conan-io/conan-center-index/issues/3183

Do you know why this isn't getting picked up?

macumber:
@jmarrec I'd like to merge this to develop, merge develop to master, then tag 1.1.1.  You ok with that?

jmarrec:
@macumber Sure thing.

PR #313, opened 2021-02-14
Remove precision from InspectorGadget, revert to default value instead of 0 when clearing

### LINKED ISSUE #270: Don't default to "0" after the user cleared an input field
Enhancement Request
--------------
When the user clears an input field, leave it empty *or* default to energyplus default value instead of writing "0".

## Detailed Description

Windows 10 Home
OpenStudioApplication 1.1.0

OpenStudioApplication should behave as energyplus when the user clears an input field (i.e. sets its value to "").
IDF Editor will leave the field empty and energyplus will use the default value when simulating.
OpenStudioApplication writes "0", thus forcing overwriting energyplus default.

Example with OS:SpaceInfiltration:DesignFlowRate "Constant Term Coefficient": the default value is "1". If you clear it and focus on another field, "" will be automatically replaced by "0". If you don't pay attention to this and don't read the simulation warnings, it will screw the whole simulation.

## Possible Implementation
Either leave it empty (i.e. leave it to energyplus) or default to energyplus default value.
If possible, an help text to explicitly show what the default value is would be a nice addition.

I don't know how easy it is to retrieve energyplus default values, thus leaving the field empty might be the easiest implementation. It is also the best option for allowing an easier maintenance of your models: in case the next version of energyplus uses another default value, you won't "benefit" from it if the current default value is hardcoded in your OSM/IDF file.

macumber:
fixes #270


jmarrec:
@macumber Have you tested inside the OSApp by clicking around?

PR #318, opened 2021-02-28
Gridview updates

macumber:
Overview:
- Minimize number of times we refresh the entire gridview.  Do not refresh to select objects.  Add new objects to bottom of grid (regardless of sorting). Disable or hide objects on deletion.
- Set stylesheets once and then control appearance with dynamic properties.  Reduces amount of time parsing QSS. Widgets update their own style when possible.
- Store widget state in member variables rather than rely on Qt methods like `isVisible` as Qt does optimizations on its functions and they may not be updated in a slot. 
- Refactor classes to make members private and reduce interdependencies between widgets.
- Add unit tests for widgets
- Add support for MicroProfiler on Windows

Remaining bugs (please add if you find any new ones):
- [x] SpaceTypesLoads, Inherited load schedules should not be deletable in inspector
- [x] SpaceTypesLoads, Drop zones keep dropped item even if set is rejected
- [x] SpaceTypesLoads, Locked/defaulted style updates for inherited schedules is odd
- [x] SpaceTypesLoads, Add tool tip on locked load definition
- [x] FacilitiesStories, Filter story by nom z works, but clear filter is not working
- [x] ThermalZoneHVAC, zone equipment not updating when add or remove
- [x] ThermalZoneHVAC, setpoint schedules not updating on remove
- [x] ThermalZoneSetpoints, apply all not immediately updating
- [x] HVACRefrigeration, duplicate case/walk-in added on drop 
- [x] ScriptsTab, leaving tab to go to Run Tab causes double delete

Deferred performance issues (will add new issues):
- [ ] SpacesSpaces, Clicking ThermalZone items in drop zones to show in inspector is very slow
- [ ] ThermalZones, Clicking ZoneHVAC items in drop zones to show in inspector is very slow
- [ ] SubTabView, disable redraws when doing bulk add or delete (e.g. delete entire story on SpacesGeneral)
- [ ] HVACRefrigeration, change to a regular grid view
- [ ] All, Look at replacing some connect and singleshot timers with Qt::ConnectionType = QueuedConnection | UniqueConnection

Partially addresses #149


src/openstudio_lib/DesignDayGridView.cpp
```diff
@@ -136,6 +132,7 @@ void DesignDayGridView::onAddClicked() {
     model::DesignDay(m_gridController->model());
   } else {
     for (auto& obj : m_gridController->selectedObjects()) {
+      // TODO: copy properties of selected to new?
```
macumber, line 135: @macumber todo

src/openstudio_lib/OSCollapsibleItemList.cpp
```diff
@@ -177,7 +177,7 @@ void OSCollapsibleItemList::addCollapsibleItem(OSCollapsibleItem* collapsibleIte
     collapsibleItem->itemList()->selectItem(collapsibleItem->itemList()->firstItem());
   }
 
-  collapsibleItem->setSelected(true);
+  //collapsibleItem->setSelected(true);
```
macumber, line 180: @macumber check on this

src/openstudio_lib/OSDropZone.cpp
```diff
@@ -513,19 +514,23 @@ void OSDropZoneItem::mouseReleaseEvent(QGraphicsSceneMouseEvent* event) {
...
   setObjectName("OSDropZone");
-
-  QString style("QWidget#OSDropZone {\
-    background: #CECECE;\
-    border: 2px dashed #808080;\
-    border-radius: 5px; }");
-  setStyleSheet(style);
+  setProperty("focused", false);
+  setStyleSheet(
+    "QWidget#OSDropZone[focused=\"true\"] { border: 2px dashed #808080; border-radius: 5px; background:#ffc627; } "
```
macumber, line 519: @macumber update with the Locked=0, Focused=0, Defaulted=0 style

src/openstudio_lib/RenderingColorWidget.cpp
```diff
@@ -62,8 +62,13 @@ RenderingColorWidget2::RenderingColorWidget2(QWidget* parent) : QWidget(parent)
   m_renderColorButton->setFixedSize(QSize(20, 20));
   layout->addWidget(m_renderColorButton);
 
-  bool isConnected = connect(m_renderColorButton, SIGNAL(clicked()), this, SLOT(renderColorButtonClicked()));
-  OS_ASSERT(isConnected);
+  connect(m_renderColorButton, &QPushButton::clicked, this, &RenderingColorWidget2::renderColorButtonClicked);
+}
+
+void RenderingColorWidget2::setLocked(bool locked) {
```
macumber, line 68: @macumber does this need styling?

src/shared_gui_components/LocalLibraryController.cpp
```diff
@@ -121,12 +121,14 @@ void LocalLibraryController::openBclDlg() {
 }
 
 LocalLibraryController::~LocalLibraryController() {
-  if (localLibraryView) {
-    delete localLibraryView;
-  }  // TODO what is this?
-
-  if (libraryView) {
-    delete libraryView;
+  if (localLibraryView && !localLibraryView.isNull()) {
```
macumber, line 124: @macumber todo

jmarrec:
I'm not sure whether it's this PR or something else, but I'm seeing UI changes between 1.1.0 and this PR.

1.1.0

image


this PR

image

jmarrec:
1.1.0

image


This PR

image

jmarrec:
This is definitely faster. I took a model with 255 surfaces, uses a stopwatch to see how much time it takes to draw the the Facility > Surfaces subtab.

1.1.0: 25 seconds
This PR: 7 seconds

jmarrec:
> ** ThermalZoneSetpoints, apply all not immediately updating**

Yes, I was going to say that. Apply all does the job but I have to go to another tab and come back to see it.

jmarrec (commented):
I barely made a dent into this review for now... will have to revisit. This is huge @macumber Thanks a ton for your hard work. Thanks for adding tests, that's great. Performance definitely improved too which is awesome. 

We'll need to fix the UI display issues though. 

(I'd like to also add benchmarks if we can, not sure exactly how yet... at least one or two going from one tab to the surface tab for eg, to see if we can use that to tweak performance and measure impact more reliably than visually)

jmarrec:
Bintray is browing out again today: https://status.bintray.com/

macumber:
Thanks for the review @jmarrec!  Yes there are some UI changes, some are intentional and others might not be.  If the new styles are better we can keep them or we can try to match what they were, I'm open.

I changed the way that styles are applied.  We now set a stylesheet once (with different styles depending on object properties) then we just change the object properties.  There is a good example in `OSDoubleEdit2::OSDoubleEdit2`.  The common properties for widgets are:

- Locked, cannot edit
- Focused, will be used as source for Apply To Selected
- Auto/Defaulted, non-user specified value

I tried to keep the styles and colors consistent for these properties between the different widgets (e.g. OSDoubleEdit2, OSDropZone2, etc).  These are pretty easy to adjust.

OSWidgetHolder is the thing that surrounds the widget, it has styles for:

- Selected, will be used as target for Apply To Selected
- Even, the alternating color for the grid

OSWidgetHolder is the thing that has the borders, if we want to get rid of those.

src/openstudio_lib/test/FacilityShading_GTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+  auto objectSelector = getObjectSelector(gridController);
+
+  processEvents();
+
+  auto modelObjects = gridController->modelObjects();
+  ASSERT_EQ(2u, modelObjects.size());
+  EXPECT_EQ(shadingGroup1.handle(), modelObjects[0].handle());
+
+  auto selectableObjectsSet = objectSelector->selectableObjects();
+  EXPECT_EQ(shadingGroups.size(), selectableObjectsSet.size());
```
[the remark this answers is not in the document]
macumber: You're right, thanks for fixing

macumber:
I saw NREL was making changes to move away from bintray, is it just a matter of updating our remotes to point at a new remote?  Is there a new public-conan and a new commercialbuilding remote?

jmarrec:
@macumber you modified the commercialbuilding remote already. As you know Bintray is sunsetting (May 1st). I modified the bincrafters remote right now by pushing to develop.

PR #321, opened 2021-03-09
Update to OpenStudio 3.2.0-alpha based on EnergyPlus v9.5.0 official

jmarrec:
I use the nightly build from today 2021-03-09. I handle the fact that OS SDK now ships a Ubuntu 18.04 and a Ubuntu 20.04. I bumped ruby to use 2.7.2 (and all conan deps from OS Sdk for that matter).
Remove two fields in ConstructionAirBoundary too.


macumber:
Included in https://github.com/openstudiocoalition/OpenStudioApplication/pull/329

PR #329, opened 2021-04-24
Update os 3 2 0 rc1

macumber:


macumber:
@jmarrec you had most of the updates to 3.2 in place, i just bumped it to the rc1.  I think we should go ahead and merge this, just wanted to wait to check with you

PR #330, opened 2021-04-28
Add benchmarking facilities to quantify performance of OSApp

jmarrec:
I've been trying to set up some benchmarking code for the past day, so we have something in place to:
* Quantify the savings the Grid view update @macumber  did #318 
* have better tools to improve it further

I think I finally spat something that works

I am benchmarking the creation of the "Spaces > Surfaces" subtab, since this is a source of major slow down since you logically have way more surfaces than spaces (A rectangular space has 6...).

Here's a visualization that shows the current develop against #318 

image


CMakeLists.txt
```diff
@@ -243,6 +243,8 @@ option(BUILD_PACKAGE "Build package" OFF)
 
 option(BUILD_TESTING "Build testing" OFF)
 
+option(BUILD_BENCHMARK "Build benchmarking targets" OFF)
```
jmarrec, line 246: new cmake option

src/openstudio_lib/test/SpacesSurfaces_Benchmark.cpp
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  state.SetComplexityN(state.range(0));
+
+}
+
+// This doesn't work, it goes [6, 8, 16...]
+//BENCHMARK(BM_SpacesSurfaces)->RangeMultiplier(2)->Range(6, 6<<6)->Unit(benchmark::kMillisecond)->Complexity();
+BENCHMARK(BM_SpacesSurfaces)->Arg(6)->Arg(12)->Arg(24)->Arg(48)->Arg(96)->Arg(192)->Arg(384)->Arg(768) // ->Arg(1536) // 1536 takes 89 secodns AFTER dan's improvements
+  ->Unit(benchmark::kMillisecond)->Complexity();
```
jmarrec, line 63: Actual parametric benchmark, with complexity calculation.

Note: In this case I'm trying to compare two very different branches, so I couldn't really do two `static void BM_xxx` functions to compare in the same, so there's just the one

PR #354, opened 2021-05-28
Internationalization: translate app (to french)

jmarrec:


src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1165,6 +1170,11 @@ void OpenStudioApp::readSettings() {
   QSettings settings(organizationName, applicationName);
   setLastPath(settings.value("lastPath", QDir::homePath()).toString());
   setDviewPath(openstudio::toPath(settings.value("dviewPath", "").toString()));
+  m_currLang = settings.value("language", "en").toString();
```
macumber, line 1173: There would be a menu item to change this on the fly right?

src/openstudio_lib/CMakeLists.txt
```diff
@@ -707,6 +707,19 @@ set(${target_name}_moc
...
+)
+#set_source_files_properties(${TS_FILES} PROPERTIES OUTPUT_LOCATION $<TARGET_FILE_DIR:${target_name}>/translations)
+
+qt5_create_translation(QM_FILES
+  ${PROJECT_SOURCE_DIR}
+  ${${target_name}_SRC}
+  ${TS_FILES})
+#include_directories(SYSTEM ${RUBY_INCLUDE_DIRS})
+
+add_custom_target(translations ALL DEPENDS ${QM_FILES})
```
macumber, line 721: Are `QM_FILES` compiled in to the app or installed as a resource?

src/openstudio_lib/DesignDayGridView.cpp
```diff
@@ -57,41 +57,41 @@
 // These defines provide a common area for field display names
 // used on column headers, and other grid widgets
 
-#define NAME "Design Day Name"
-#define SELECTED "All"
+#define NAME openstudio::DesignDayGridView::tr("Design Day Name")
```
macumber, line 60: Why can't you just do:

`#define NAME tr("Design Day Name")`

?

src/openstudio_lib/LocationTabView.cpp
```diff
@@ -782,7 +773,8 @@ void LocationView::setCalendarYear(int year) {
 void LocationView::setFirstDayofYear(const QString& firstDayofYear) {
   m_yearDescription->resetCalendarYear();
 
-  m_yearDescription->setDayofWeekforStartDay(firstDayofYear.toStdString());
+  auto idx = YearSettingsWidget::validDayofWeekforStartDay().indexOf(firstDayofYear);
```
macumber, line 776: Good catch

src/openstudio_lib/YearSettingsWidget.cpp
```diff
@@ -510,55 +497,65 @@ void YearSettingsWidget::onDstEndDayWeekMonthChanged() {
...
-  std::vector<std::string> result;
+QStringList YearSettingsWidget::validDayofWeekforStartDay() {
 
-  std::map<int, std::string> nameMap = DayOfWeek::getNames();
-
-  for (auto it = nameMap.begin(); it != nameMap.end(); ++it) {
-    result.push_back(it->second);
-  }
+  // Similar to model::YearDescription::validDayofWeekforStartDayValues(), but we need to be able to translate them
+  QStringList result  = {
```
macumber, line 516: Translating from strings in the IDD is going to be expensive.  I wonder if we could write a script to export all the object names and enums from the IDD, translate that, then just feed `tr` the strings from the IDD?

macumber: We could use https://translate.google.com/?sl=auto&tl=en&op=docs to do a head start?  Or maybe people prefer to not translate the IDD items?

macumber:
Looks good!

src/openstudio_lib/DesignDayGridView.cpp
jmarrec: Because when the macro is expanded, it doesn't know the context of the Q_Object it's doing it for. See https://doc.qt.io/qt-5/i18n-source-translation.html#using-tr-for-all-literal-text

src/openstudio_lib/CMakeLists.txt
jmarrec: Initially I was embeddeding them in the `openstudio.qrc`, but it's more trouble that needed, and now that I carry over some of the Qt internal ones, I figured we'd just install them where Qt expects them, that is `PREFIX_PATH/bin/translations` (Content/MacOS/blablalba on mac, I haven't played much with this)

jmarrec: see `qt.conf` file too.

src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -241,9 +245,9 @@ void OpenStudioApp::onMeasureManagerAndLibraryReady() {
       while (!measureManager().isStarted()) {
         LOG(Fatal, "Failed to start the Measure Manager on try " << currentTry << ", timeout reached.");
         QMessageBox msgBox;
-        msgBox.setWindowTitle("Timeout");
+        msgBox.setWindowTitle(tr("Timeout"));
         msgBox.setIcon(QMessageBox::Critical);
-        msgBox.setText("Failed to start the Measure Manager. Would you like to retry?");
+        msgBox.setText(tr("Failed to start the Measure Manager. Would you like to retry?"));
```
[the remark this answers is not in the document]
jmarrec: Qt Linguist stores not only the original stirng, but also the filepath, class, and line number. So it should be able to update it just fine. Defining a variable wouldn't be terribly useful in this context.

lupdate/lrelease is ran via cmake on each build, so it stay current. From personal experience with this, you should commit (or at least stage) the OpenStudioApp_xx.ts file often, as when modifying a large number of lines (or worse, messing with the CMake settings regarding qt5_create_translations) it could sometimes have some hiccups.

src/openstudio_lib/YearSettingsWidget.cpp
jmarrec: I'd say generally I'm going to be opposed to translating the IDD Enums. For days and months it makes a ton of sense, but otherwise not so much. It's not like you're going to find the E+ I/O reference guide in anything else than English, so I'd want to keep that aligned.

jmarrec:
I have no clue how CI is picking up a circular dependency issue. I have built this locally on both Ubuntu and macOS without any issue.

```
[0/2] Re-checking globbed directories...
ninja: error: dependency cycle: src/openstudio_lib/translations -> src/openstudio_lib/CMakeFiles/translations -> src/openstudio_lib/OpenStudioApp_fr.qm -> src/openstudio_lib/.lupdate/OpenStudioApp_fr.ts.stamp -> src/openstudio_app/__/shared_gui_components/moc_BusyWidget.cpp -> Products/libopenstudio_lib.a -> src/openstudio_lib/CMakeFiles/openstudio_lib.dir/ApplyMeasureNowDialog.cpp.o -> cmake_object_order_depends_target_openstudio_lib -> src/openstudio_lib/translations
```

luisclv:

jmarrec:
recheck

kentakahas:

PR #367, opened 2021-05-31
366 geometry editor map

macumber:


src/openstudio_lib/OSWebEnginePage.cpp
```diff
@@ -33,9 +33,25 @@
...
+  info.setHttpHeader("Accept-Language", "en-US,en;q=0.9,es;q=0.8,de;q=0.7");
+  info.setHttpHeader("Access-Control-Allow-Origin", "*");
+}
+
+OSWebEnginePage::OSWebEnginePage(QObject* parent) : QWebEnginePage(parent) {
+  OSUrlRequestInterceptor* interceptor = new OSUrlRequestInterceptor(this->profile());
+  this->profile()->setUrlRequestInterceptor(interceptor);
+}
+
+OSWebEnginePage::~OSWebEnginePage() {}
```
jmarrec, line 53: I think you need to free the OSUrlRequestInterceptor*. You should perhaps just use a unique_ptr?

jmarrec: https://doc.qt.io/qt-5/qwebengineprofile.html#setUrlRequestInterceptor

> https://doc.qt.io/qt-5/qwebengineprofile.html#setUrlRequestInterceptor

I interpret that as meaning it doesn't free.

jmarrec: welllll I guess you do pass this as a parent : `OSUrlRequestInterceptor* interceptor = new OSUrlRequestInterceptor(this);`...

PR #388, opened 2021-06-23
Internationalization: Translate to Japanese

kentakahas:
Adding the Japanese translation for the first tab of OpenStudio. Some phrases may sound a little off since the wording is different from English.


jmarrec:
@kentakahas Thanks a bunch! I definitely cannot comment on the translation itself, but I did check in qt linguist and noticed all green checkmarks, which was super nice, so looks good to me!

jmarrec (commented):
all green checkmarks in Qt Linguist

PR #390, opened 2021-06-26
Fix issue with large models

### LINKED ISSUE #378: Models with a large number of Spaces or ThermalZones crashes the application
[2021.06.08_Code_Model.zip](https://github.com/openstudiocoalition/OpenStudioApplication/files/6626622/2021.06.08_Code_Model.zip)

Issue overview
--------------
In a model with a relatively large number of Spaces (and Thermal Zones) makes the window not respond, and crashes when changing to the ThermalZones or Spaces tab.

## Current Behavior
The window stops responding, and then closes.

## Expected Behavior
The tab should open.

## Steps to Reproduce
Create a new model with 500 spaces

```ruby
require 'openstudio'
path = ENV['HOME'] + '/Desktop/test.osm'
model = OpenStudio::Model::Model.new
for a in 1..500
  z = OpenStudio::Model::ThermalZone.new(model)
  s = OpenStudio::Model::Space.new(model)
  s.setThermalZone(z)
end

model.save(path,true)
```

The model is handled fine in the OS App V1.1.0. This approach does not crash the application, but shows an empty tab.

I attached the model in question, but I do not think it is model dependent, the crash might be related to the attached model having surfaces, equipment, etc.

## Possible Solution

Some memory issue with Qt maybe? 

### Environment

 - Platform (Operating system, version): Windows 10 Pro v2004 19041.985
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): OpenStudio App V1.2.0

### Context
I cannot continue working in the model, I cannot see ZoneHVACEquipment or apply Thermostats.

macumber:
Fixes #378


src/shared_gui_components/OSGridView.cpp
```diff
@@ -78,6 +78,8 @@
 #  define HEIGHT 60
 #endif
 
+const int NUM_ROWS_PER_GRIDLAYOUT = 5;
```
macumber, line 81: This is set low right now for testing, I will probably bump it to 51 for real use.

jmarrec:
```suggestion
constexpr int NUM_ROWS_PER_GRIDLAYOUT = 5;
```

src/shared_gui_components/OSGridView.cpp
```diff
@@ -364,6 +370,34 @@ void OSGridView::recreateAll() {
   }
 }
 
+int OSGridView::layoutIndex(int row) const {
+  return row / NUM_ROWS_PER_GRIDLAYOUT;
+}
+
+int OSGridView::rowInLayout(int row) const {
+  return row % NUM_ROWS_PER_GRIDLAYOUT;
+}
```
jmarrec, line 379:
```suggestion
constexpr int OSGridView::layoutIndex(int row) const {
  return row / NUM_ROWS_PER_GRIDLAYOUT;
}

constexpr int OSGridView::rowInLayout(int row) const {
  return row % NUM_ROWS_PER_GRIDLAYOUT;
}
```

src/shared_gui_components/OSGridView.cpp
jmarrec: if I recall correctly, the original design was to have 50 rows per grid layout. Did you change that when you refactored?

Anyhoo, if we're displaying X by X, is there any way we can do that multithreaded?

macumber:
@jmarrec I don't think we can create QWidgets in another thread, they have to be in the main thread.  I don't quite understand why but that is what the documentation seems to say:

https://doc.qt.io/qt-5/threads-qobject.html

src/shared_gui_components/OSGridView.cpp
macumber: I don't think this can be constexpr because row is not known at compile time

jmarrec: The constexpr specifier declares that it is possible to evaluate the value of the function or variable at compile time. Which it can.

```c++
#include <iostream>

constexpr int NUM_ROWS_PER_GRIDLAYOUT = 51;

constexpr int layoutIndex(int row) {
  return row / NUM_ROWS_PER_GRIDLAYOUT;
}

int main() {
  int number;
  std::cout << "Please enter a row: ";
  std::cin >> number;

  int idx = layoutIndex(number);
  std::cout << "layoutIndex=" << idx << '\n';
  return idx;
}
```

```
$ g++ -std=c++17 main.cpp -o test && ./test
Please enter a row: 109
layoutIndex=2
```

jmarrec: Anyways, that's just pedantism on my part :)

macumber: image

src/openstudio_app/CMakeLists.txt
```diff
@@ -132,6 +132,9 @@ add_executable(${target_name}
 
 if (NINJA)
   target_compile_definitions(${target_name} PRIVATE NINJA=1)
+elseif(WIN32)
+  # increase stack size
+  target_link_options(${target_name} PRIVATE /STACK:8000000)
```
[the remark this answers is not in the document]
jmarrec: @macumber Should we make it 8388608?

jmarrec (approved):

src/openstudio_app/CMakeLists.txt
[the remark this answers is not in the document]
jmarrec: Aside from that, this can merge

PR #391, opened 2021-06-26
Update to OS 3.2.1

macumber:


jmarrec:
@macumber good. We need to update the OSMs though

jmarrec:
I updated the OSMs, will let CI churn, and if it's green we can merge. Thanks Dan.

PR #405, opened 2021-07-22
Implement simple search in geometry preview

### LINKED ISSUE #404: Geometry preview - search feature
Enhancement Request
--------------
Use case: an EnergyPlus simulation is run and there are some errors / warnings about surfaces / thermal zones in `eplusout.err`. It can be useful to quickly show the concerned objects in geometry preview. 

This is more relevant for larger buildings or when no naming convention was applied (ie `Surface 202`, `Surface 203` ...), as it'll often imply long look up times.

antoine-galataud:
Fixes #404


macumber:
I would like to merge the version on https://github.com/openstudiocoalition/OpenStudioApplication/pull/409 instead

PR #406, opened 2021-07-28
Add Clang tidy file

jmarrec:
Add some clang tidy file to the project. The goal is add more diagnostics and clean up our existing code base which has some non negligible legacy.

I've modified a couple of files while tweaking the clang-tidy config, to see where we would stand and what impact it may have.


src/model_editor/AccessPolicyStore.cpp
```diff
@@ -131,26 +128,28 @@ bool AccessParser::startElement(const QString& /*namespaceURI*/, const QString&
...
       }
     }
-    if (fieldName.size() && accessRule.size()) {
+    if (!fieldName.isEmpty() && !accessRule.isEmpty()) {
       AccessPolicy::ACCESS_LEVEL level = AccessPolicy::FREE;
-      if (!accessRule.compare("locked", Qt::CaseInsensitive)) {
+      if (accessRule.compare("locked", Qt::CaseInsensitive) == 0) {
         level = AccessPolicy::LOCKED;
-      } else if (!accessRule.compare("hidden", Qt::CaseInsensitive)) {
+      } else if (accessRule.compare("hidden", Qt::CaseInsensitive) == 0) {
```
macumber, line 152: Did clang tidy make this change or did you have to?

src/openstudio_lib/LocationTabView.cpp
```diff
@@ -799,25 +791,25 @@ void LocationView::setDaylightSavingsTime(bool enabled) {
...
 void LocationView::setDstEndDayOfWeekAndMonth(int newWeek, int newDay, int newMonth) {
-  model::RunPeriodControlDaylightSavingTime dst = m_model.getUniqueModelObject<model::RunPeriodControlDaylightSavingTime>();
+  auto dst = m_model.getUniqueModelObject<model::RunPeriodControlDaylightSavingTime>();
 
   dst.setEndDate(NthDayOfWeekInMonth(newWeek), DayOfWeek(newDay), MonthOfYear(newMonth));
 }
 
 void LocationView::setDstEndDate(const QDate& newdate) {
-  model::RunPeriodControlDaylightSavingTime dst = m_model.getUniqueModelObject<model::RunPeriodControlDaylightSavingTime>();
+  auto dst = m_model.getUniqueModelObject<model::RunPeriodControlDaylightSavingTime>();
```
macumber, line 812: Is this just preferring to use auto?

src/.clang-tidy
```diff
@@ -0,0 +1,59 @@
+---
```
macumber, line 1: Would there be a github actions to call this?

src/model_editor/AccessPolicyStore.cpp
jmarrec: It did for me. QString::compare returns an int.

(Also, aside from int -> bool conversion, `!str.compare("other")` to express "if it matches" is HELLA confusing to me)

src/openstudio_lib/LocationTabView.cpp
jmarrec: auto if template / cast: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-auto.html#cast-expressions

I'm not a proponent of (almost) always auto, but `T a = mo.cast<T>()` or `T a = mo.getUniqueModelObject<T>()` is a good candidate.

src/.clang-tidy
jmarrec: In an ideal world / down the line, we would **enforce** it via a github actions workflow yes. For now, let's not:
* There are a gazillion warnings right now that would need to be eliminated before we do, some of which do not have an automatic fix
* The .clang-tidy config file is not at a point where I feel confident every option enable is something we want

so instead I propose we develop with clang-tidy enabled (you can have your favorite editor including MSVC do code analysis with clang-tidy, I set it up for gViM myself), and take it from there, as in: when touching a file for another purpose, look at the warnings/fixes clang-tidy proposes and maybe fix them or adjust the .clang-tidy config if you feel they aren't good enough.

Once we have a decent feeling that the config is what we want, then we can do a bulk `--fix` (well, using the `run-clang-tidy.py` script on the compile database to be exact) on the entire code base, and see how many warnings we're left dealing with manually.

As a side note, clang-tidy is extensible, meaning you can write your own checks with fixits. You can benefit from the clang AST matchers to do bespoke refactorings (and it's something I have toyed with recently) that aren't as dumb as just using regexes. I made an example for you to look at if you're curious: https://godbolt.org/z/8ffYPrr3z

jmarrec: Another note is that cppcheck is definitely still useful: see how https://github.com/openstudiocoalition/OpenStudioApplication/pull/406/checks?check_run_id=3182084911 caught some mistakes I've done here?

src/model_editor/AccessPolicyStore.cpp
macumber: Seems like a good change

src/.clang-tidy
macumber: I finally figured out how to run this in Visual Studio, seems like a good approach to start with

macumber (approved):

PR #409, opened 2021-08-01
Fix 404

macumber:
https://github.com/openstudiocoalition/OpenStudioApplication/pull/405


macumber:
@antoine-galataud what do you think about the changes on this branch?

PR #424, opened 2021-08-26
Try new logo (I may have missed a few files)

jmarrec:


macumber:
One other file that should be updated is `\src\openstudio_lib\images\image_header.png`, its shown on modal dialogs

jmarrec:
@macumber I think I may have messed up the debian post installation script for Ubuntu.

https://github.com/openstudiocoalition/OpenStudioApplication/blob/6896b020c0203596f6f4209e66d8e6db4a2cd762/debian/postinst.in#L52

The path to copy it to might need changing, that shouldn't be to the mimetype folder but the application one. We'll need to test that on on a clean machine and one that had a previous version installed, to see if the new icon shows correctly in both situations.

PR #448, opened 2021-10-26
445 inspector gadget all fields

### LINKED ISSUE #445: Inspector does not all access to all fields
Issue overview
--------------
When inspecting an object in the inspector, fields that are not yet created are not editable.  Previously, we could adjust the minimum number of fields in the OS IDD to create all fields.  Since that is no longer an option, suggest that we create all fields when loading an object in the inspector.

## Current Behavior
fueltype

## Expected Behavior
All fields should be editable

jmarrec:
Supersedes #446 

- Fix #445 
- I took the liberty of completely cleaning the entire code of the InspectorGadget while I was at it. (except the use of a static initializer and the const_cast stuff, that I left untouched). see the first commit + review comments to see the actual non-cleaning changes

All screenshots below are from the 'Example Model' if you want to reproduce @macumber 

Design choices:

- A field that is meant to be locked keeps displaying as usual: purple, locked. Eg: Building Story 'Nominal Floor to Ceiling Height' (when viewed outside of the Facility tab)

image

- A field that is meant to be FREE (editable), is displayed in blue but no longer locked. Eg: Thermal Zone's & Sizing:Zone, when viewed from the Space tab

image

image

Editing a field initializes the field(s), but you have to step outside of the object and reload it to see your changes.


macumber (approved):
I added on commit to rebuild if the number of fields increases.  If you are ok with that then this can be merged.

jmarrec:
@macumber The only question I had about calling onWorkspaceObjectChanged() manually to rebuild is why it's needed in the first place? Why does the onWorkspaceObjectChange not take care of it? https://github.com/openstudiocoalition/OpenStudioApplication/blob/7c48b113ce74cb4854a71fc00948d0813dd4f57e/src/model_editor/InspectorGadget.cpp#L1089-L1102

But now I see we disconnect then reconnect the signals in between setting the value (which I find counter intuitive but it's probably for a good reason).

Anyways, I'm ok to merge this.

PR #465, opened 2021-11-14
Fixes #464

### LINKED ISSUE #464: Why is ThermalZone gridview adding ZoneControl:Humidistats
Issue overview
--------------
If I'm understanding this correctly (*not guaranteed!*), the ThermalZones tab will add ZoneControl:Humidistat objects to the zones just by displaying the 'Humidifying/Dehumidifying Setpoint Schedule" columns: 

https://github.com/openstudiocoalition/OpenStudioApplication/blob/fcd926fb70d2dc1590847677e08f585973998fdf/src/openstudio_lib/ThermalZonesGridView.cpp#L553-L555

@macumber WHY?
They are empty objects, without setpoint schedules applied, but still annoyingly break any measure logic that checks for Humidistat objects as an indication of, e.g., humidity control. I don't see why this was added - before [this commit](https://github.com/openstudiocoalition/OpenStudioApplication/commit/2fbc5b9e25fd4f831a04b838daeafb4bd9b9c9e9) the humidistat objects weren't created until a setpoint schedule was dragged to the drop zone, and that was good functionality!

## Current Behavior
Create a model with some zones using the SDK, open the model in the App and navigate to the Thermal Zones tab. Save the model and a ZoneControl:Humidistat will have been created and assigned to each zone.

## Expected Behavior
The UI should not create objects just as a consequence of displaying the model content.

 - Version of OpenStudioApplication (if using an intermediate build, include SHA): 1.2.1-rc1+b1fdf35e7f

macumber:


src/openstudio_lib/OSDropZone.hpp
```diff
@@ -126,7 +135,7 @@ class OSDropZone2
   boost::optional<OtherModelObjects> m_otherObjects;
   boost::optional<model::ModelObject> m_modelObject;
   boost::optional<model::ModelObject> m_getterResult;
-  std::vector<model::ModelObject> m_otherModelObjects;
+  std::set<model::ModelObject, ModelObjectHandleLess> m_otherModelObjects;
```
jmarrec, line 138: Were you forced to specify a custom comparator?

src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -383,35 +383,32 @@ void ThermalZonesGridController::addColumns(const QString& /*category*/, std::ve
...
         }
+        return result;
       });
 
       boost::optional<std::function<void(model::ThermalZone*)>> resetCoolingSchedule([](model::ThermalZone* z) {
         if (boost::optional<model::ThermostatSetpointDualSetpoint> thermostat = z->thermostatSetpointDualSetpoint()) {
           thermostat->resetCoolingSetpointTemperatureSchedule();
+          if (!thermostat->heatingSetpointTemperatureSchedule()) {
+            z->resetThermostatSetpointDualSetpoint();
+          }
```
jmarrec, line 402: ok, if you remove the cooling schedule and no heating schedule was assigned, you reset the ThermostatSetpointDualSetpoint.

src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -383,35 +383,32 @@ void ThermalZonesGridController::addColumns(const QString& /*category*/, std::ve
...
 
       boost::optional<std::function<bool(model::ThermalZone*)>> isDefaulted;
 
-      boost::optional<std::function<std::vector<model::ModelObject>(model::ThermalZone*)>> otherObjects([](model::ThermalZone* z) {
+      boost::optional<std::function<std::vector<model::ModelObject>(const model::ThermalZone*)>> otherObjects([](const model::ThermalZone* z) {
         std::vector<model::ModelObject> result;
-        boost::optional<model::Thermostat> thermostat = z->thermostat();
-        boost::optional<model::ThermostatSetpointDualSetpoint> thermostatDualSetpoint = z->thermostatSetpointDualSetpoint();
-        if (thermostat) {
+        if (boost::optional<model::ThermostatSetpointDualSetpoint> thermostat = z->thermostatSetpointDualSetpoint()) {
```
jmarrec, line 410: True that at this point the only possibility of DualSetpoint (and Thermostat is an base class without a public Ctor, so effectively an abstract one)

src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -430,35 +427,32 @@ void ThermalZonesGridController::addColumns(const QString& /*category*/, std::ve
...
+          if (!thermostat->coolingSetpointTemperatureSchedule()) {
+            z->resetThermostatSetpointDualSetpoint();
+          }
         }
       });
 
       boost::optional<std::function<bool(model::ThermalZone*)>> isDefaulted;
 
-      boost::optional<std::function<std::vector<model::ModelObject>(model::ThermalZone*)>> otherObjects([](model::ThermalZone* z) {
+      boost::optional<std::function<std::vector<model::ModelObject>(const model::ThermalZone*)>> otherObjects([](const model::ThermalZone* z) {
```
jmarrec, line 452: Ok now I understand why these are now `const T *`. We aren't instantiating children object anymore, so const correctness was applied.

(PS: This syntax is really gross)

src/openstudio_lib/ThermalZonesGridView.cpp
```diff
@@ -470,42 +464,40 @@ void ThermalZonesGridController::addColumns(const QString& /*category*/, std::ve
...
 
       boost::optional<std::function<void(model::ThermalZone*)>> resetHumidifyingSchedule([](model::ThermalZone* z) {
-        if (boost::optional<model::ZoneControlHumidistat> thermostat = z->zoneControlHumidistat()) {
-          thermostat->resetHumidifyingRelativeHumiditySetpointSchedule();
-          if (!thermostat->dehumidifyingRelativeHumiditySetpointSchedule()) {
-            thermostat->remove();
+        if (boost::optional<model::ZoneControlHumidistat> humidistat = z->zoneControlHumidistat()) {
+          humidistat->resetHumidifyingRelativeHumiditySetpointSchedule();
+          if (!humidistat->dehumidifyingRelativeHumiditySetpointSchedule()) {
+            z->resetZoneControlHumidistat();
```
jmarrec, line 489: I had a doubt about this. But this does remove it too, so no problem: https://github.com/NREL/OpenStudio/blob/db1d522abee3d173e904f5c78a716c6274dfb092/src/model/ThermalZone.cpp#L1179-L1185

jmarrec (approved):

src/openstudio_lib/OSDropZone.cpp
```diff
@@ -929,4 +913,43 @@ boost::optional<model::ModelObject> OSDropZone2::updateGetterResult() {
   return m_getterResult;
 }
 
+void OSDropZone2::updateOtherModelObjects() {
```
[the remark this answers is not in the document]
macumber: Hey Julien, I agree.  I added `otherObjects` to the `DropZoneConcept` when I refactored the grid view code.  The `otherObjects` function tells the drop zone to watch some other object (in this case the thermostat or humidistat) for changes and refresh the drop zone if that other object changes in addition to changes in the main object (in this case the zone).  The problem here was that the return value of `otherObjects` changes when the thermostat/humidistat are removed/added.  Previously, drop zone only called `otherObjects` at bind time (which is why I created thermostat/humidistat at bind time in my refactor).  Now, the drop zone calls `otherObjects` at each refresh.  I made the `otherObjects` function take a `const T*` because if you change the main object inside the `otherObjects` function you will get an infinite loop of calling `refresh` then `otherObjects`.

src/openstudio_lib/OSDropZone.hpp
macumber: I just wanted to make sure that we were using the handle as the set key, I wasn't sure what `std::set<IdfObject>` would use as the key on it's own.  Using a set just simplified some of the code inside for adding or removing objects to watch (we only want to watch each other object once).

src/openstudio_lib/ThermalZonesGridView.cpp
macumber: Totally agree, drop zone is way more complicated than I would like.

src/openstudio_lib/ThermalZonesGridView.cpp
macumber: Yeah, I checked that too.  Just removing the humidistat does not trigger change signals from the zone object.  I guess now that we are properly watching the humidistat all the time we could go back to the old way because we would refresh when the humidistat was removed.  Let me know if you want me to change that?  It would protect us from future breakages if OpenStudio ever changes `resetZoneControlHumidistat` to not remove the thermostat.

src/openstudio_lib/FacilityShadingGridView.cpp
```diff
@@ -503,7 +503,7 @@ void FacilityShadingGridController::addColumns(const QString& category, std::vec
           CastNullAdapter<model::ShadingSurface>(&model::ShadingSurface::setConstruction),
           boost::optional<std::function<void(model::ShadingSurface*)>>(NullAdapter(&model::ShadingSurface::resetConstruction)),
           boost::optional<std::function<bool(model::ShadingSurface*)>>(NullAdapter(&model::ShadingSurface::isConstructionDefaulted)),
-          boost::optional<std::function<std::vector<model::ModelObject>(model::ShadingSurface*)>>(), DataSource(allShadingSurfaces, true));
+          boost::optional<std::function<std::vector<model::ModelObject>(const model::ShadingSurface*)>>(), DataSource(allShadingSurfaces, true));
```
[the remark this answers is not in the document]
macumber: I made the `otherObjects` method take a `const T*` to remind us not to change the main object in that call.  The `getter` and `isDefaulted` functions should also take a `const T*` but that would have been a much larger change so I skipped it.

PR #471, opened 2021-11-29
Add more logging around saving files and handle case where save as fails

macumber:


src/openstudio_lib/OSDocument.cpp
```diff
@@ -1444,14 +1468,18 @@ bool OSDocument::saveAs() {
...
+    if (saved) {
+      LOG(Debug, "Saving " << modelPath << " complete");
+      this->setSavePath(toQString(modelPath));
+      this->markAsUnmodified();
+      fileSaved = true;
+    } else {
+      LOG(Debug, "Saving " << modelPath << " failed");
+      QMessageBox::warning(this->mainWindow(), tr("Failed to save model"),
+                           tr("Failed to save model, make sure that you do not have the location open and that you have correct write access."));
+    }
```
jmarrec, line 1482: Noted, that's the actual change that might have an impact. The rest is just `LOG(Debug, ...)`

jmarrec (approved):

PR #474, opened 2021-12-11
Update to use release cert and add rc1

macumber:


macumber:
Idea is to merge this to develop, merge https://github.com/openstudiocoalition/OpenStudioApplication/pull/473 to master, then create tag v1.3.0-rc1.  That is all we need to do to create a build with the release cert right?

.github/workflows/app_build.yml
```diff
@@ -536,7 +536,7 @@ jobs:
           -CIUserToken "${{ secrets.SIGNPATH_CI_TOKEN }}" `
           -OrganizationId "97f757f1-cd69-467b-b87b-db3eb5102a57" `
           -ProjectSlug "OpenStudioApplication" `
-          -SigningPolicySlug "No_Approval_GHA" `
+          -SigningPolicySlug "No_Approval_Release_GHA" `
```
macumber, line 539: @jmarrec I created this signing policy

jmarrec: :+1:

PR #490, opened 2022-02-20
Increase stack size on Windows CI builds

macumber:
Stack size was not getting increased on Windows CI because Ninja generator was used


macumber:
Apply fix from #378 for CI builds

PR #496, opened 2022-03-22
Additional output to run tab

### LINKED ISSUE #110: Hard to see errors generated in ModelToIdf job in App
You have to go to the tree, right click the job and do show errors and warnings.  Seems like these should get printed to the screen as well

2-28-2014 2-14-52 pm

### LINKED ISSUE #437: No error message for failed simulation with empty spaces
Issue overview
--------------

> The problem here was that there were spaces or zones without anything in them in the OSM file. After finding and deleting them it ran again. Can also be the cause when error file just stops abruptly. Would be nice to add a check in the initializing or translating to EP.

Reported at https://unmethours.com/question/52769/openstudio-fails-without-stating-errors-found-and-blank-eplusouterr-file/

## Expected Behavior
An error message should be shown in the Run Tab describing the error

macumber:
Add additional arguments to the run command, needs OS 3.4.0.  We should consider removing the socket and just use stdout.

Fixes #110
Fixes #437

### DIFF
--- a/src/openstudio_lib/RunTabView.cpp
+++ b/src/openstudio_lib/RunTabView.cpp
@@ -230,6 +230,9 @@ void RunView::playButtonClicked(bool t_checked) {
     auto workflowJSONPath = QString::fromStdString(workflowPath.string());
     QStringList arguments;
     arguments << "run"
+              << "--show-stdout"
+              << "--style-stdout"
+              << "--add-timings"
               << "-s" << QString::number(m_runTcpServer->serverPort()) << "-w" << workflowJSONPath;
 
     LOG(Debug, "openstudioExePath='" << toString(openstudioExePath) << "'");


jmarrec:
I resolved the merge conflicts, will let CI run and give it a test then merge.

jmarrec:
I tried the installer from https://github.com/openstudiocoalition/OpenStudioApplication/pull/496/commits/9ccc0b7fe094f82f0f56bd3a5747e080be5ae49a. The simulation did not run and the OSApp looked "just fine"  when really the CLI, which is 3.3.0 on this one and does not yet have these new `--show-stdout` and co flags, was failing. 

I think we should really check exit code / exit status on https://github.com/openstudiocoalition/OpenStudioApplication/blob/b286ab82c8e446b5a055d5465a82e4a5a3821755/src/openstudio_lib/RunTabView.cpp#L181-L197

jmarrec:
You can't pass both `--socket `and `--show-stdout`. `--show-stdout` which I added in 3.4.0 is simply opening a socket to retrieve the data from openstudio-workflow-hem, which is **exactly** what we're already doing here already.

jmarrec:
I revamped the PR:

macumber:
It looks like you moved this to https://github.com/openstudiocoalition/OpenStudioApplication/pull/513, is that right?

PR #502, opened 2022-04-24
Fixes #494

### LINKED ISSUE #494: Opening simulation run directory fails
Issue overview
--------------
User reported issue here:
https://unmethours.com/question/70637/unable-run-simulation-openstudio/

## Current Behavior
Simulation directory does not open, critical error shown

## Expected Behavior
Simulation directory opens

## Possible Solution
May need to set the URL to be in TolerantMode
https://doc.qt.io/qt-5/qdesktopservices.html#openUrl
https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/openstudio_lib/RunTabView.cpp#L177

### Environment

macumber:
- Fixes #494

### DIFF
--- a/src/openstudio_lib/RunTabView.cpp
+++ b/src/openstudio_lib/RunTabView.cpp
@@ -171,9 +171,9 @@ RunView::RunView() : QWidget(), m_runSocket(nullptr) {
 
 void RunView::onOpenSimDirClicked() {
   std::shared_ptr<OSDocument> osdocument = OSAppBase::instance()->currentDocument();
-  QString url = QString::fromStdString((getCompanionFolder(toPath(osdocument->savePath())) / toPath("run")).string());
-  QUrl qurl = QUrl::fromLocalFile(url);
-  if (!QDesktopServices::openUrl(qurl)) {
+  path runDir = getCompanionFolder(toPath(osdocument->savePath())) / toPath("run");
+  QString path = QDir::toNativeSeparators(toQString(runDir));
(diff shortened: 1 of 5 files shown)


src/openstudio_lib/RunTabView.cpp
```diff
@@ -171,9 +171,9 @@ RunView::RunView() : QWidget(), m_runSocket(nullptr) {
 
 void RunView::onOpenSimDirClicked() {
   std::shared_ptr<OSDocument> osdocument = OSAppBase::instance()->currentDocument();
-  QString url = QString::fromStdString((getCompanionFolder(toPath(osdocument->savePath())) / toPath("run")).string());
-  QUrl qurl = QUrl::fromLocalFile(url);
-  if (!QDesktopServices::openUrl(qurl)) {
+  path runDir = getCompanionFolder(toPath(osdocument->savePath())) / toPath("run");
+  QString path = QDir::toNativeSeparators(toQString(runDir));
```
macumber, line 175: FYI @j-lorrey : the issue here is that the OpenStudio path class on Windows uses UTF-16 encoding, we were passing that to `QString::fromStdString` which expects a UTF-8 string.  The fix is to use [toQString](https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/model_editor/Utilities.cpp#L77) to convert the path, that method handles conversion from UTF-16 on Windows.

jmarrec (approved):

PR #497, opened 2022-04-01
filter scene by air loop

antoine-galataud:
Fixes #413

PR #510, opened 2022-04-29
render 3D scene on demand

### LINKED ISSUE #509: Geometry preview - on-demand rendering
Enhancement Request
--------------

Makes 3D rendering of geometry more CPU-efficient by using on-demand rendering and not continuous animation.

## Detailed Description

The main implication will be on the fluidity of scene rendering upon user interactions: with continuous animation there's no `render()` call on each change, so it may appear fluid. On the other hand continuous animation eats up the CPU, and can make the whole experience jerky (possibly related with https://github.com/openstudiocoalition/OpenStudioApplication/issues/118)

## Possible Implementation

The change consists in disabling continuous animation and rendering only when something changes in the scene (ie user changing camera position, or clicking an option in the menu). Some testing required on several devices to ensure it doesn't hurt user experience.

antoine-galataud:
Fixes #509

### DIFF
--- a/src/openstudio_lib/library/geometry_preview.html
+++ b/src/openstudio_lib/library/geometry_preview.html
@@ -2288,11 +2288,13 @@ function init(os_data_in) {
   orthographicControls.minDistance = 10;
   orthographicControls.maxDistance = 1000;
   orthographicControls.enabled = true;
+  orthographicControls.addEventListener('change', requestRenderIfNotRequested);
 
   perspectiveControls = new THREE.OrbitControls(perspectiveCamera, renderer.domElement);
   perspectiveControls.minDistance = 10;
   perspectiveControls.maxDistance = 1000;
   perspectiveControls.enabled = true;
+  perspectiveControls.addEventListener('change', requestRenderIfNotRequested);
 
   // initialize camera and controls
   setAllCameraAngles(-30, 30, 0);
@@ -2538,6 +2540,7 @@ function onDocumentMouseClick(event) {
     } else {
       removeSelection();
     }
+    requestRenderIfNotRequested();
   }
 }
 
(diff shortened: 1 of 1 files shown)


jmarrec:
@antoine-galataud Thanks for the PR. Could you make a a small `peek` gif showing before/after to ease in reviewing please?

antoine-galataud:
@jmarrec I had prepared a peek capture for this, but actually there's actually no observable difference before and after the change. If you're still interested I can produce one that also includes CPU metrics (ie htop summary in a corner)

antoine-galataud:
Here are 2 screen captures (large model). It's actually slower and less fluid before the change, due to the constant load of animation + recording of the video. Notice also the large diff in CPU load at the beginning (when no user interaction happens).

Before:

before_animate

After:
after_no_animate

jmarrec:
tested locally, did not notice any difference. @macumber wanna take a pass before I merge?

src/openstudio_lib/library/geometry_preview.html
```diff
@@ -2538,6 +2540,7 @@
     } else {
       removeSelection();
     }
+    requestRenderIfNotRequested();
```
macumber, line 2543: Do we need to do this on changing render mode, hiding surfaces, or other things too?

macumber:
Left one comment, I haven't had time to build and try on my own yet.  My other comment is I'd like to make sure all of these improvements make it back to https://github.com/NREL/openstudio-common-measures-gem/ as well.  That way we don't diverge from those measures too much.

Thanks again @antoine-galataud !

src/openstudio_lib/library/geometry_preview.html
antoine-galataud: Interactions on the dat.gui menu trigger an `onChange` event that is captured by `update()`. I added a call to `requestRenderIfNotRequested` there too (see https://github.com/openstudiocoalition/OpenStudioApplication/pull/510/files#diff-574d82d3d9fea756c6e5daf9d279da971c8ad80e9171c4f1ff33de92938dfbfcR3070).

I think it covers the needs, but maybe you think of other interactions that should request rendering?

antoine-galataud:
@macumber @jmarrec I can handle adding #510 and #497 to https://github.com/NREL/openstudio-common-measures-gem/ once they are both validated and merged here. Just let me know.

src/openstudio_lib/library/geometry_preview.html
macumber: Gotcha!

macumber (approved):

jmarrec:
@antoine-galataud That'd be great if you could upstream to common measure gem! Thanks!

Issue #230, opened 2020-09-09
Remove explicit constructor cppcheck suppression


macumber:
https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/.github/workflows/cppcheck.yml

jmarrec:
Fixed via #511

PR #511, opened 2022-04-29
[#230] Remove explicit constructor cppcheck suppression

j-lorrey:


j-lorrey:

macumber:
If you look at the cppcheck failures https://github.com/openstudiocoalition/OpenStudioApplication/runs/6236087647?check_suite_focus=true you will see a list off all the lines that have this error, e.g.: `[src/bimserver/ProjectImporter.hpp:56]:(style),[noExplicitConstructor],Class 'ProjectImporter' has a constructor with 1 argument that is not explicit.`

You then just need to change `ProjectImporter(QWidget* parent);` to `explicit ProjectImporter(QWidget* parent);`

See https://en.cppreference.com/w/cpp/language/explicit for discussion of what explicit means/does

macumber:
Your changes in https://github.com/openstudiocoalition/OpenStudioApplication/pull/511/commits/a4d88e7025d570e26f527913d4dacd3f41639c34 look mostly right, but I'm not sure why you had to do a force push?  Using force push is a last resort thing, you typically don't want to do it.

The remaining C++ compile errors are things were we might need to change the code back to allow implicit casting and suppress the explicit constructor warning for individual lines.  Is that something you want to try?  It does get a bit deep with the template errors.

jmarrec:
Apparently the force push is because the commit was broken due to an extra explicit keyword in the cpp file for ProjectImporter. I actually disagree that force pushing is a last ressort if you're doing it on your own PR branch, and I personally do it all the time (to squash commits, rebase, etc) in order to avoid having a branch with 100 commits like "fix typo", "do another".
Force pushing to develop/master is another story

src/shared_gui_components/OSConcepts.hpp
```diff
@@ -261,7 +261,12 @@ std::function<RetType(FromDataType*, Param1)> ProxyAdapter(RetType (ToDataType::
 class ConceptProxy
 {
  public:
+  // TODO: This can't be explicit with boost::optional,
+  // see https://github.com/openstudiocoalition/OpenStudioApplication/pull/511#pullrequestreview-963081285
+  // I do not know how to get around it, so will cppcheck-suppress it (and let future us worry about moving to std::optional)
+  // Please increment counter next time: **Hours wasted trying to understand the issue: 2.5 **
   template <typename T>
+  // cppcheck-suppress noExplicitConstructor
```
jmarrec, line 269: Admission of failure for the win!

jmarrec:
@macumber @j-lorrey I'll let you read up my monologue above and take another pass, then please go ahead and merge

jmarrec:
I'm merging, we can revisit if need be

src/shared_gui_components/OSConcepts.hpp
macumber: That is the smart move

PR #513, opened 2022-05-02
Option to enable Verbose output to run tab + styled fallback when no socket

### LINKED ISSUE #110: Hard to see errors generated in ModelToIdf job in App

### LINKED ISSUE #437: No error message for failed simulation with empty spaces

jmarrec:
* Fixes #110
* Fixes #437

* Add a "verbose output" checkbox to the run tab
* This is saved in the user preferences
* When enabled, this will do openstudio --verbose run -w workflow.osw, and style the stdout/stderr
* When no socket can be opened, the fallback will also be styled similarly to when sockets are used
    * This cannot be the default behavior though (I'm keeping the explicit socket first in the line) because during the E+ sims the updates don't come directly, there may be a stdout flush missing in the openstudio_cli.rb   

Verbose_simulation_output


src/openstudio_lib/RunTabView.cpp
```diff
@@ -231,19 +273,33 @@ void RunView::playButtonClicked(bool t_checked) {
...
 
     QStringList arguments;
-    if (haveConnection) {
+    LOG(Debug, "Checkbox is checked? " << std::boolalpha << m_verboseOutputBox->isChecked());
+    if (m_verboseOutputBox->isChecked()) {
+      arguments << "--verbose";
+    } else {
+      // If not verbose, we save the stdout/stderr to a file, like historical
+      // Actually, we don't, we just read it
+      // m_runProcess->setStandardOutputFile(toQString(stdoutPath));
```
macumber, line 287: Will E+ stdout and stderr still be in files in the sim dir for the user to read later?  I see stdout-energyplus and stderr-energyplus files but I'm not entirely sure which process was writing those, the app or the cli.

src/openstudio_lib/RunTabView.cpp
```diff
@@ -257,30 +313,28 @@ void RunView::playButtonClicked(bool t_checked) {
     m_state = State::stopped;
     m_textInfo->clear();
 
-    if (haveConnection) {
-      m_progressBar->setMinimum(0);
-      m_progressBar->setMaximum(State::complete);
-      m_progressBar->setValue(0);
-    } else {
-      m_progressBar->setMinimum(0);
-      m_progressBar->setMaximum(0);
```
macumber, line 266: Setting the maximum to 0 here was to use this feature of the progress bar "If minimum and maximum both are set to 0, the bar shows a busy indicator instead of a percentage of steps. "  This was because I wasn't able to update `m_state` without the socket.  It doesn't look like you are updating `m_state` in the stdout reader so I would suggest leaving this as is.

macumber: Oh, I see that you are doing that, nevermind this is a good improvement

macumber (approved):
Looks good, just want to make sure the user still has stdout and stderr in files for later

src/openstudio_lib/RunTabView.cpp
jmarrec: stdout-energyplus stderr-energyplus are written by workflow gem.

What I am no longer outputing is the model_companion_dir/stdout and stderr files. I see no value in keeping those, when we are seeing them live. I don't see any use case for seeing them afterwards, but perhaps there is one. In which case the RunView::readyReadStandardOutput and RunView::readyReadStandardError should be writting to these files in append mode in addition of displaying them live. thoughts @macumber ?

macumber: Ok yeah, I don't think those two files are useful.  Thanks Julian!

jmarrec: cf 4411071f97974765f3d84eee9d272998ad1a17b0

PR #193, opened 2020-07-21
Add new field for SimulationControl: `DoHVACSizingSimulationforSizingPeriods` and `MaximumNumberofHVACSizingSimulationPasses`

jmarrec:
Fix #192 - Add new field for SimulationControl: `DoHVACSizingSimulationforSizingPeriods` and `MaximumNumberofHVACSizingSimulationPasses`

PR #515, opened 2022-05-03
Fix #477 - SubSurface types don't line up with many subsurfaces in space

### LINKED ISSUE #477: SubSurface types don't line up with many subsurfaces in space
Issue overview
--------------
With many subsurfaces in a space, the subsurface types don't line up well

image

This makes it hard to edit subsurfaces

## Steps to Reproduce
See OSM in issue files repo

jmarrec:
* Fix #477

### DIFF
--- a/src/openstudio_lib/SpacesSubsurfacesGridView.cpp
+++ b/src/openstudio_lib/SpacesSubsurfacesGridView.cpp
@@ -399,7 +399,8 @@ void SpacesSubsurfacesGridController::addColumns(const QString& category, std::v
                               boost::optional<std::function<bool(model::SubSurface*)>>(), DataSource(allSubSurfaces, true));
       } else if (field == SUBSURFACETYPE) {
         addComboBoxColumn<std::string, model::SubSurface>(
-          Heading(QString(SUBSURFACETYPE)), static_cast<std::string (*)(const std::string&)>(&openstudio::toString),
+          Heading(QString(SUBSURFACETYPE)),
+          std::function<std::string(const std::string&)>(static_cast<std::string (*)(const std::string&)>(&openstudio::toString)),
           std::function<std::vector<std::string>()>(&model::SubSurface::validSubSurfaceTypeValues),
           CastNullAdapter<model::SubSurface>(&model::SubSurface::subSurfaceType),
           CastNullAdapter<model::SubSurface>(&model::SubSurface::setSubSurfaceType),
--- a/src/openstudio_lib/SpacesSubsurfacesGridView.cpp
+++ b/src/openstudio_lib/SpacesSubsurfacesGridView.cpp
@@ -407,7 +407,7 @@ void SpacesSubsurfacesGridController::addColumns(const QString& category, std::v
             CastNullAdapter<model::SubSurface>(&model::SubSurface::resetSubSurfaceType)),  // New since 3.1.0
           boost::optional<std::function<bool(model::SubSurface*)>>(
             CastNullAdapter<model::SubSurface>(&model::SubSurface::isSubSurfaceTypeDefaulted)),  // New since 3.1.0
-          boost::optional<DataSource>(allSubSurfaces, true));
+          DataSource(allSubSurfaces, true));
       } else if (field == MULTIPLIER) {
         addValueEditColumn(Heading(QString(MULTIPLIER)), NullAdapter(&model::SubSurface::multiplier), NullAdapter(&model::SubSurface::setMultiplier),
                            boost::optional<std::function<void(model::SubSurface*)>>(NullAdapter(&model::SubSurface::resetMultiplier)),


src/openstudio_lib/SpacesSubsurfacesGridView.cpp
```diff
@@ -399,15 +399,16 @@ void SpacesSubsurfacesGridController::addColumns(const QString& category, std::v
...
+          std::function<std::string(const std::string&)>(static_cast<std::string (*)(const std::string&)>(&openstudio::toString)),
           std::function<std::vector<std::string>()>(&model::SubSurface::validSubSurfaceTypeValues),
           CastNullAdapter<model::SubSurface>(&model::SubSurface::subSurfaceType),
           CastNullAdapter<model::SubSurface>(&model::SubSurface::setSubSurfaceType),
           boost::optional<std::function<void(model::SubSurface*)>>(
             CastNullAdapter<model::SubSurface>(&model::SubSurface::resetSubSurfaceType)),  // New since 3.1.0
           boost::optional<std::function<bool(model::SubSurface*)>>(
             CastNullAdapter<model::SubSurface>(&model::SubSurface::isSubSurfaceTypeDefaulted)),  // New since 3.1.0
-          boost::optional<DataSource>(allSubSurfaces, true));
+          DataSource(allSubSurfaces, true));
```
jmarrec, line 411: The actual minimum fix is this line..

macumber: What is this doing?  I'd guess calling a different overload but all the overloads take a boost::optional<DataSource>?

jmarrec: It's hard to tell, isn't it? The bowels of the OSGridController / OSConcepts is too clever...

I started with a more complicated solution which involved add a new `addComboBoxColumn` overload in OSGridController. Then I tested every combination and remove stuff to realize that this was the actual minimum fix and left it there as it was late. But the fact that you are curious now just justified nerd-snipping myself so I looked...

When you pass `boost::optional<DataSource>(allSubSurfaces, true)`, you jump directly to 
https://github.com/openstudiocoalition/OpenStudioApplication/blob/14a08a83ec4443d7531516938bff0678578caa36/src/shared_gui_components/OSGridController.hpp#L236-L246

When you pass `DataSource(allSubSurfaces, true)`, you go to https://github.com/openstudiocoalition/OpenStudioApplication/blob/14a08a83ec4443d7531516938bff0678578caa36/src/shared_gui_components/OSGridController.hpp#L213-L223 when then goes to the one above.

The main difference is actually the type of the `choice` func.

```
std::function<std::vector<ChoiceType>()>
std::function<std::vector<ChoiceType>(DataSourceType*)>
```

Somehow, the choices is broken in the original and ends up messing up the display.

What I don't get is how in the original it accepts `std::function<std::vector<std::string>()>(&model::SubSurface::validSubSurfaceTypeValues)`as a valid function for choices when `std::function<std::vector<ChoiceType>(DataSourceType*)> choices` is what is expected in that overload.

jmarrec: Scratch that, actually both go through the same chain of two addComboBox!

The issue is how t_source is gotten. When passing `DataSource(allSurfaces, true)`,  the DataSource has m_wantsPlaceholder = true

When trying to pass `boost::optional<DataSource>(allSubSurfaces, true)`,  m_wantsPlaceholder = false

jmarrec: The issue is that, if wanting to explicitly say boost::optional<DataSource> it should have been written as

```c++
boost::optional<DataSource>(DataSource(allSubSurfaces, true))
```

Here, it's picking up a boost::optional<DataSource>::optional overload that takes an additional bool, so what it sees is

```diff
-boost::optional<DataSource>(   DataSource(allSubSurfaces, true)     )
+boost::optional<DataSource>(  DataSource(allSubSurfaces)              ,      true)
```

https://www.boost.org/doc/libs/1_77_0/libs/optional/doc/html/boost_optional/reference/header__boost_optional_optional_hpp_/header_optional_optional_values.html#reference_operator_template

    template<class... Args> explicit optional ( in_place_init_t, Args&&... args ) ; R

    template<class... Args> explicit optional ( in_place_init_if_t, bool condition, Args&&... args )

macumber: Dear lord, I think we have one too many overloads

macumber: Also maybe a good option to move to std::optional.

jmarrec: The Overlord of overloads. Moving to `std::` from boost in general would be nice down the line, but will be painful.
There are often subtle and sometimes less subtle differences, like some functions would behave differently or some methods are not available in STL

std::optional is probably going to be about fine.

std::filesystem very much less so: linking to it is a PITA, and there are a few things that are fundamentally broken IMHO, like the way it handles trailing separators (see this compiler explorer I made on the subject to showcase it to someone else: https://godbolt.org/z/G57vzfoEW). Funnily enough, as I have started to port the SDK to boost 1.78, I realized the boost::filesystem::path is now using V4 api which is equally dumb with trailing separators to align on STL...

https://www.boost.org/users/history/version_1_78_0.html

> [Filesystem](https://www.boost.org/libs/filesystem/):

>    v4: path::filename and path::iterator no longer return an implicit trailing dot (".") element if the path ends with a directory separator. Instead, an empty path is returned, similar to C++17 std::filesystem. This also affects other methods that are defined in terms of iterators or filename, such as path::stem, path::compare or lexicographical_compare. For example, path("a/b/") == path("a/b/.") no longer holds true. ([#193](https://github.com/boostorg/filesystem/issues/193))
>    v4: path::lexically_normal no longer produce a trailing dot (".") element and omits a directory separator after a trailing dot-dot ("..") element in the normalized paths.
 >   v4: path append operations now consider root name and root directory in the appended path. If the appended path is absolute, or root name is present and differs from the source path, the resulting path is equivalent to the appended path. If root directory is present, the result is the root directory and relative path rebased on top of the root name of the source path. Otherwise, the behavior is similar to v3. ([#214](https://github.com/boostorg/filesystem/issues/214))
>    path::lexically_normal now converts directory separators to preferred separators in the root name of the normalized paths.

PR #517, opened 2022-05-04
#479 - Find outdated measures in your local BCL folder compared to Online BCL

### LINKED ISSUE #479: Measure updates on BCL don't work
Issue overview
--------------
Download a measure from BCL, update measure on BCL, app detects new version is available but won't download it.

## Current Behavior
App detects new version is available but won't download it.
image

## Expected Behavior
App should show check box next to measures with update and allow user to download them.

jmarrec:
- Fix #479
- Find outdated measures in your local BCL folder compared to Online BCL

Kapture 2022-05-05 at 01 16 24


src/openstudio_lib/OSDocument.cpp
```diff
@@ -1747,6 +1747,7 @@ void OSDocument::openMeasuresBclDlg() {
 }
 
 void OSDocument::on_closeMeasuresBclDlg() {
+  OSAppBase::instance()->checkForRemoteBCLUpdates();
```
macumber, line 1750: Wouldn't it be better to check for this in `openMeasuresBclDlg` so you see updated measures when you get there?  Does the update take too long to do there?

macumber: Oh nice, I see you added a button, that is even better!

src/shared_gui_components/MeasureManager.cpp
```diff
@@ -945,18 +945,43 @@ bool MeasureManager::checkForUpdates(const openstudio::path& measureDir, bool fo
...
+      remoteBCL.updateMeasures();
+    }
+  }
+}
+
 void MeasureManager::downloadBCLMeasures() {
-  auto remoteBCL = new RemoteBCL();
-  int numUpdates = remoteBCL->checkForMeasureUpdates();
+  RemoteBCL remoteBCL;
+  int numUpdates = remoteBCL.checkForMeasureUpdates();
```
macumber, line 975: Bizarre that we only called this here

macumber:
I guess a future enhancement would be to allow selectively updating certain measures from the measure dialog, that would be a new enhancement though.  This is great.

macumber (approved):

src/openstudio_lib/OSDocument.cpp
jmarrec: I have been wondering what the best UX would be, like perhaps when you first go to the scripts tab it tells you whether you have outdated remote BCL measures in your local BCL **and** checks if you have project measures outdated compared to my measures dir, or whether you could it on click of a button.

src/shared_gui_components/MeasureManager.cpp
jmarrec: Also bizarre it's actually never called anywhere. There's a lot of cruft in the codebase, cleaning it up at some point would be nice...

jmarrec:
> I guess a future enhancement would be to allow selectively updating certain measures from the measure dialog, that would be a new enhancement though. This is great.

a future enhancement might be to adjust where the trigger is, and perhaps the UI and do something similar to the "Sync project measures" dialog.

As far as being able to individually select measures, I'm not sure what the use case would be. IMHO your local ~/BCL folder should be kept in sync with the remote BCL. If you need to have another version, then make a copy to your local ~/OpenStudio/Measures directory, no?

PR #518, opened 2022-05-05
Bump SDK to v3.4.0 official

jmarrec:


jmarrec:
I changed the cache key, it was still picking up rc2: https://github.com/openstudiocoalition/OpenStudioApplication/runs/6307391418?check_suite_focus=true#step:17:49

we should revisit whether caching the build dir is worth it or not...

PR #525, opened 2022-05-06
Workaround for #521: download the missing vcomp140.dll

### LINKED ISSUE #521: Need to install vcomp140.dll for EnergyPlus on Windows
Issue overview
--------------

Need to install vcomp140.dll for EnergyPlus on Windows, this is a new DLL: https://github.com/NREL/EnergyPlus/pull/9310

This is really an OpenStudio SDK issue (which is my fault, when I added it to OpenStudio's install rules, I made a typo, so it's actually not added to the installer), yet we probably don't want to ship a broken install so we should work around it.

### Environment

 - Platform (Operating system, version): Windows
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): v1.4.0-rc1

### Context

cf #519

jmarrec:
- Fix #521

### DIFF
--- a/FindOpenStudioSDK.cmake
+++ b/FindOpenStudioSDK.cmake
@@ -231,3 +231,16 @@ else()
     )
   endif()
 endif()
+
+# TODO: Temporary workaround for missing vcomp140.dll in SDK package for EnergyPlus
+if(WIN32)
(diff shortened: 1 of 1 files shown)


jmarrec:
https://github.com/openstudiocoalition/OpenStudioApplication/runs/6323222983?check_suite_focus=true#step:16:203

Confirmed that the artifact on windows does have vcomp140.dll in the EnergyPlus/ folder after extracting the zip.

PR #533, opened 2022-05-10
#521 - Udpate SHA for OS SDK 3.4.0 on Windows

jmarrec:
cf #521, part 2
OS SDK binaries for v3.4.0 have been hot swapped to include the missing dll

Update SHA and revert workaround in abc8bf0c65ee9e833ae72232a0c6054a37463f4c (PR #525)

### DIFF
--- a/.github/workflows/app_build.yml
+++ b/.github/workflows/app_build.yml
@@ -292,7 +292,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: build/
-        key: ${{ matrix.os }}-build-cache-${{ secrets.CACHE_KEY }}
+        key: ${{ matrix.os }}-build-cache-v1-${{ secrets.CACHE_KEY }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/app_build.yml
```diff
@@ -292,7 +292,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: build/
-        key: ${{ matrix.os }}-build-cache-${{ secrets.CACHE_KEY }}
+        key: ${{ matrix.os }}-build-cache-v1-${{ secrets.CACHE_KEY }}
```
macumber, line 295: I think you can just change the cache key secret right?

macumber:
Nice!

.github/workflows/app_build.yml
jmarrec: Yeah I forgot  about that, thanks for the reminder

PR #537, opened 2022-05-12
Start at adjusting dependencies to match what windeployqt suggests

### LINKED ISSUE #358: Cannot run geometry editor on systems without OpenGL 2.0
Issue overview
--------------
Geometry Editor fails to run on systems without OpenGL 2.0

## Current Behavior
App will either crash or (on VirtualBox VMs) graphics will display inverted when going to geometry tab

## Expected Behavior
Geometry tab should work

## Steps to Reproduce
Run OpenStudio in Windows VM on VirtualBox with enable 3D graphics option off

## Possible Solution
Ship Angle OpenGL dlls (d3dcompiler_47.dll, libEGL.dll, libGLESv2.dll) with application

macumber:
Fixes #358


macumber:
windeployqt is also suggesting:

- iconengines (dir)
- imageformats (dir)
- networkinformation (dir)
- position (dir)
- tls (dir)
- Qt6SerialPort.dll

Those all seem plausible (except maybe Qt6SerialPort) but I am not sure if it is too heavy.  Total extra size is 2.67 MB so maybe best to just add them.

macumber:
I'll also want to check this on Mac, I think Qt6::DBus might need to be in `QT_LIBS`

macumber:
Updated SHA to include hotfix for https://github.com/NREL/OpenStudio/issues/4584

FindOpenStudioSDK.cmake
```diff
@@ -19,7 +19,7 @@ if(APPLE)
     set(OPENSTUDIO_EXPECTED_HASH 5ef630a7282b10c8a9bba34f00539a98)
     set(OPENSTUDIO_PLATFORM "Darwin-arm64")
   else()
-    set(OPENSTUDIO_EXPECTED_HASH a5f16ff19c3a36f5a44fd0153da21633)
+    set(OPENSTUDIO_EXPECTED_HASH 5a1e5fdfc61a879a9d72dcf625a83e65)
```
macumber, line 22: @j-lorrey @jmarrec 

This includes hotfix for https://github.com/NREL/OpenStudio/issues/4584, not sure if arm64 on Mac was affected by that issue?

macumber:
@jmarrec I think this is good to go if it looks ok to you

FindOpenStudioSDK.cmake
jmarrec: arm64 already had libintil8, since I have noticed it was missing during my own testing.

src/openstudio_app/CMakeLists.txt
```diff
@@ -177,6 +177,11 @@ if( APPLE )
     VERBATIM
   )
 
+  #add_custom_command(TARGET ${target_name}
+  #  POST_BUILD
+  #  COMMAND ${QT_MACDEPLOY_QT} $<TARGET_BUNDLE_DIR:${target_name}> -verbose=2
+  #)
```
jmarrec, line 183: Should we use macdeployqt, and linuxdeployqt?

CMakeLists.txt
```diff
@@ -521,7 +521,7 @@ message(STATUS "QT_INSTALL_DIR=${QT_INSTALL_DIR}")
 
 find_package(Qt6 ${QT_VERSION} COMPONENTS CoreTools GuiTools WidgetsTools QmlTools WebEngineCoreTools REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
 
-find_package(Qt6 ${QT_VERSION} COMPONENTS Core Core5Compat Gui Widgets Sql Network Xml Concurrent PrintSupport Quick QuickWidgets Qml WebChannel Positioning WebEngineCore WebEngineWidgets REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
+find_package(Qt6 ${QT_VERSION} COMPONENTS Core Core5Compat Gui Widgets Sql Svg Network Xml Concurrent PrintSupport Quick QuickWidgets Qml WebChannel Positioning WebEngineCore WebEngineWidgets REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
```
jmarrec, line 524: We aren't using QSvg nor do we have any .svg files in the repo. Why would we need it?

jmarrec (commented):
https://github.com/openstudiocoalition/OpenStudioApplication/blob/5094cc0c160cda287f0854d24e91d50515dad583/src/openstudio_app/CMakeLists.txt#L252-L254

What you're doing is aligning Mac/Linux onto what windeployqt (that we already use) is doing right?

src/openstudio_app/CMakeLists.txt
macumber: Eventually I think so.  When I ran macdeployqt the resulting package crashed, I don't know why.

CMakeLists.txt
macumber: I am not 100% sure, Qt might be using SVG internally, I think some of our web assests use SVG.  The app does load that dll when it is installed so I figured it was good to just include it:

image

FindOpenStudioSDK.cmake
macumber: Nice thanks!

CMakeLists.txt
jmarrec: I just checked out the source code for windeployqt, what it does is to find some DLLs you do use, then it reads what this DLL depends on and includes them as well. (findDependentQtLibraries -> readExecutable -> readPeExecutable)

src/openstudio_app/CMakeLists.txt
```diff
@@ -350,10 +355,22 @@ if( NOT APPLE )
...
     endforeach()
 
+    list(APPEND QT_DIRS iconengines)
+    list(APPEND QT_DIRS imageformats)
+    list(APPEND QT_DIRS networkinformation)
+    list(APPEND QT_DIRS position)
+    list(APPEND QT_DIRS tls)
+    foreach(DIR ${QT_DIRS})
+      install(DIRECTORY $<TARGET_FILE_DIR:${target_name}>/${DIR} DESTINATION bin/ COMPONENT OpenStudioApp)
+    endforeach()
```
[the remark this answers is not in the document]
jmarrec: Oh I see your comment above, "Those all seem plausible (except maybe Qt6SerialPort) but I am not sure if it is too heavy. Total extra size is 2.67 MB so maybe best to just add them."

PR #540, opened 2022-06-03
Style updates

### LINKED ISSUE #536: Text too large on OSItems
Enhancement Request
--------------
The text on OSItems seems larger than it used to be

## Detailed Description
Comparison of 1.4.0 vs 1.3.0

image

## Possible Implementation
Adjust stylesheet for OSItem

macumber:
Fixes #536


src/openstudio_app/StartupView.cpp
```diff
@@ -139,7 +139,7 @@ StartupView::StartupView(QWidget* parent) : QWidget(parent), m_templateListModel
   cancelButton->setStyleSheet("QPushButton { font: bold; }");
 
   auto chooseButton = new QPushButton();
-  chooseButton->setObjectName("StandardBlueButton");
+  chooseButton->setObjectName("StandardGrayButton");
```
jmarrec, line 142: StartupView is dead code actually... I realized that today. Its not used anywhere.

https://github.com/openstudiocoalition/OpenStudioApplication/blob/f36960c500b4d32de8f9a5a174f0ac563f9ef744/src/openstudio_app/OpenStudioApp.cpp#L926-L936

jmarrec:
Notes: testing on mac

* I find the Schedules Set view less readable, seems to center text vertically, where before it would show the start of the schedule name
* On the Schedules view, the Y axis (schedule value) tick marks are gone.
* I prefered the style of the "Add new dday/holiday rule" before, where the add button was blue

Schedules


* Constructions > Constructions: the default MainRightColumnController is "Edit" (which is blank) whereas it was "My model" before which made more sense. Edit: This is actually true on all subtabs anywhere (eg: Schedules > Schedules Sets)

* I prefer the style of the subtabs buttons (blue, rounded) that you changed, thanks for that.
image

* The "Supply / Demand Equipment" labels on the HVAC Loop view may have become too small. I prefered it before

HVAC_loop


* There isn't a change here, but I think  maybe we should restyle the loading / progress bar: 

image

jmarrec:
> * Constructions > Constructions: the default MainRightColumnController is "Edit" (which is blank) whereas it was "My model" before which made more sense. Edit: This is actually true on all subtabs anywhere (eg: Schedules > Schedules Sets)

Actually, it's probably just because it "remembers" the subtab of the mainrightcolumn controller even if you switch tab, which wasn't something I had realized before. Is that newish?

src/openstudio_lib/GridItem.cpp
```diff
@@ -1292,9 +1295,9 @@ void SystemCenterItem::paint(QPainter* painter, const QStyleOptionGraphicsItem*
     painter->drawRect(0, yOrigin, m_hLength * 100, 100);
   }
 
-  QFont f;
-  f.setPixelSize(20);
-  painter->setFont(f);
+  QFont font = painter->font();
+  font.setPointSize(14);
+  painter->setFont(font);
```
jmarrec, line 1300: I find that too small on my own machine.

macumber: Yeah, I think these are going to be really hard to get perfect because everyone has different opinions + different screen sizes.  I agree that slightly larger text is a good idea for the Supply/Demand Equipment labels.

macumber:
> Actually, it's probably just because it "remembers" the subtab of the mainrightcolumn controller even if you switch tab, which wasn't something I had realized before. Is that newish?

If I am following correctly, you mean you go to Schedule Sets tab, select Edit (which doesn't make sense), then go to Construction Sets and Edit is still active (does not go back to My Model).  Is that what you mean?  I don't know if that used to be different, it acts the same in version 1.0.0 which is as far back as I checked.  It kind of makes sense because if you are pulling stuff from the library in one tab you might be doing the same thing in other tabs?

src/openstudio_lib/GridItem.cpp
macumber: Here are the differences between 1.2.0 (Qt5) and 1.4.0 (Qt6) on my laptop.  I have my laptop on the recommend settings for my display, resolution 1920x1080 and app scaling 125%.  The app scaling changed a bunch between Qt5 and Qt6, the icons look all low resolution (which is why I automatically upscaled them using imagemagick, svgs would be better).

https://user-images.githubusercontent.com/1327850/205473277-04126457-9b3c-485d-bf3d-15c34dda2b2a.mp4

macumber:
The text on the OSItem is top justified for me (see the "Office Misc Occ But Really Long For Some Reason" schedule).  I do see some weird artifacts where the border around the OSItem doesn't connect all the way.  It looks that way in 1.4.0 as well.

image

src/openstudio_lib/GridItem.cpp
macumber: Here is the text at 20 pt on my computer, it gets cropped off.  I could try to find what is cropping it and extend instead.

image

macumber:
We are missing mini-icons for the UtilityBill objects, just an FYI

macumber:
@jmarrec I got my three PRs as far as I could this weekend.  If you want to review them further before merging we can, or we can merge and put out an RC.  I probably missed some stuff in this PR, I didn't go back through the entire app yet.

jmarrec:
I'm merging. Thanks for all these changes. We'll keep an eye out for this stuff during RC testing.

jmarrec:
> > Actually, it's probably just because it "remembers" the subtab of the mainrightcolumn controller even if you switch tab, which wasn't something I had realized before. Is that newish?
> 
> If I am following correctly, you mean you go to Schedule Sets tab, select Edit (which doesn't make sense), then go to Construction Sets and Edit is still active (does not go back to My Model). Is that what you mean? I don't know if that used to be different, it acts the same in version 1.0.0 which is as far back as I checked. It kind of makes sense because if you are pulling stuff from the library in one tab you might be doing the same thing in other tabs?

Yeah I think that does make sense. I just didn't realize this was the case before :)

PR #544, opened 2022-06-07
Fix #535 - Combobox offset when using fractional scaling

### LINKED ISSUE #535: Offset drop downs for measure arguments on the measure tab
Issue overview
--------------
Using a measure with drop down arguments, selecting the drop-down results in the choice list showing offset

## Current Behavior
image

## Expected Behavior
Drop down choices should appear related to the argument box
### Environment

 - Platform (Operating system, version): Windows 10
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): v1.4.0-rc1

### Context
Confusing

jmarrec:
* FIx #535

**TODO**: investigate side effects / display glitches this brings?

This only happens when you use **Fractional** Scaling, eg 250%. It works at 200% or 300%.

See this MCVE: https://github.com/jmarrec/QtTestbed/tree/main/Test-Combobox-Popup-Issue

And the bug report I filed at Qt's issue tracker: https://bugreports.qt.io/browse/QTBUG-104073


src/openstudio_app/main.cpp
```diff
@@ -215,7 +215,8 @@ int main(int argc, char* argv[]) {
     openstudio::OpenStudioApp app(argc, argv);
     openstudio::Application::instance().setApplication(&app);
 
-    app.setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor);
+    // cf #535
+    app.setHighDpiScaleFactorRoundingPolicy(Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
```
jmarrec, line 219: @macumber Do you remember why you had picked RoundPreferFloor (or actually, explicit set it, since PassThrough is the default)?
https://github.com/openstudiocoalition/OpenStudioApplication/blame/f3d7afb29064b40d75ac8c3c7ec6ca02d48b18b3/src/openstudio_app/main.cpp#L218

jmarrec:
image

jmarrec:
I replaced the gradient image (1px wide) + image header (icon + pixelised text) by the icon, a QLinearGradient and some text placed by the QPainter:

Before:

image


After:

image

jmarrec:
Final result with Muli-Regular embedded (that's what's used on the https://openstudiocoalition.org/)

image

src/shared_gui_components/OSDialog.cpp
```diff
@@ -159,14 +161,29 @@ void OSDialog::paintEvent(QPaintEvent* event) {
 
   painter.setRenderHint(QPainter::Antialiasing);
 
-  QImage centerHeader = QImage(":/images/header-backgnd-1px-wide.png");
```
macumber, line 162: Awesome, we should nuke this png at some point

src/openstudio_app/main.cpp
macumber: I remember experimenting with the different options and thinking that one was the best but whatever reason wouldn't outweigh the offset issue

PR #552, opened 2022-06-22
Fix #551 - Allow connecting a VRF Terminal from "MyModel" to a VRF System

### LINKED ISSUE #551: Add ZoneHVAC:TerminalUnit:VariableRefrigerantFlow to outside air system equipment
Enhancement Request
--------------

I would like to be able to add the ZoneHVAC:TerminalUnit:VariableRefrigerantFlow object to an outside air system.  It appears that currently this object can only be added as a zone terminal.

This will become more important as VRF systems are modeled because they are becoming more integrated with the ventilation portions of the systems (with dedicated VRF coils to temper the ventilation air).

jmarrec:
- Fix #551 
- Allow connecting a VRF Terminal from "MyModel" to a VRF System and adjust display of VRF COntroller when connected to AirLoopHVAC

551_VRF

@Ski90Moo

### DIFF
--- a/src/openstudio_lib/VRFController.cpp
+++ b/src/openstudio_lib/VRFController.cpp
@@ -49,6 +49,7 @@
 #include <openstudio/model/ComponentData_Impl.hpp>
 
 #include "../model_editor/Utilities.hpp"
+#include <boost/optional.hpp>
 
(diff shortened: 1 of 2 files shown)


jmarrec:
@macumber I'm going to merge, this has been open since mid june.

PR #553, opened 2022-06-24
Geometry preview - fix incorrect variable scope

antoine-galataud:
Backport of https://github.com/NREL/openstudio-common-measures-gem/pull/113


jmarrec:
thanks @antoine-galataud

PR #564, opened 2022-09-21
Add tab tracking with google analytics

macumber:


src/openstudio_lib/AnalyticsHelper.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+namespace openstudio {
+
+AnalyticsHelper::AnalyticsHelper(QObject* parent) : QObject(parent) {
+  m_networkAccessManager = new QNetworkAccessManager(this);
+  m_networkAccessManager->setAutoDeleteReplies(true);
+}
+
+AnalyticsHelper::~AnalyticsHelper() {
+  // cancel all futures
```
jmarrec, line 49: `delete m_networkAccessManager;` ?

src/openstudio_lib/AnalyticsHelper.cpp
```diff
@@ -0,0 +1,79 @@
...
+  query.addQueryItem("dp", path);
+  query.addQueryItem("dt", title);
+  query.addQueryItem("cd", title);
+  query.addQueryItem("an", "OSApp");
+  query.addQueryItem("av", version);
+  query.addQueryItem("z", QString::number(std::time(0)));
+
+  QByteArray body = query.query().toUtf8();
+  m_networkAccessManager->post(request, body);
+}
```
jmarrec, line 77: Should we capture subtabs? (eg: Loads on the Space Types tab)

src/openstudio_lib/MainWindow.hpp
```diff
@@ -205,6 +219,8 @@ class MainWindow : public QMainWindow
 
   QString m_lastPath;
 
+  QString m_analyticsId;
```
jmarrec, line 222: Why not a bool?

src/openstudio_lib/MainWindow.cpp
```diff
@@ -303,10 +317,41 @@ bool MainWindow::verboseOutput() const {
...
+    reply = QMessageBox::question(this, tr("Allow Analytics"),
+                                  tr("Allow OpenStudio Coalition to collect anonymous usage statistics to help improve the OpenStudio Application?"),
+                                  QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
+    if (reply == QMessageBox::Yes) {
+      emit enableAnalytics(true);
+    } else {
+      emit enableAnalytics(false);
+    }
+  }
+}
```
jmarrec, line 343: :+1: Confirmed that we do get the prompt only once, until the settings are written

src/openstudio_lib/MainWindow.cpp
```diff
@@ -285,12 +294,17 @@ void MainWindow::writeSettings() {
...
+  settings.setValue("analyticsId", m_analyticsId);
 }
 
 QString MainWindow::lastPath() const {
   return QDir().exists(m_lastPath) ? m_lastPath : QDir::homePath();
 }
 
+bool MainWindow::allowAnalytics() const {
+  return !m_analyticsId.isEmpty() && m_analyticsId != "DISABLED";
+}
```
jmarrec, line 306: Ok so that's why you're not using a bool, because you would actually need a tristate boolean or an `optional<bool>` to check whether initialized or not.

src/openstudio_lib/MainWindow.cpp
```diff
@@ -303,10 +317,41 @@ bool MainWindow::verboseOutput() const {
...
+      emit enableAnalytics(true);
+    } else {
+      emit enableAnalytics(false);
+    }
+  }
+}
+
+void MainWindow::toggleAnalytics(bool allowAnalytics) {
+  if (allowAnalytics) {
+    m_analyticsId = openstudio::toQString(openstudio::createUUID());
```
jmarrec, line 347: Ah no ok, that's why `m_analyticsId` is QString! It stores a unique id. Gotcha, makes sense!

src/openstudio_lib/AnalyticsHelper.cpp
jmarrec: Also :+1: I confirmed (by adding an `std::cout << body.toStdString() << '\n'` here) that when I tick the checkbox to not allow analytics that not analytics is actually posted.

src/openstudio_lib/MainWindow.cpp
```diff
@@ -303,10 +317,41 @@ bool MainWindow::verboseOutput() const {
...
   m_verboseOutput = verboseOutput;
 }
 
+void MainWindow::promptAnalytics() {
+  if (m_analyticsId.isEmpty()) {
+    QMessageBox::StandardButton reply;
+    // TODO: link to a privacy page on the website, explain why we need metrics
+    reply = QMessageBox::question(this, tr("Allow Analytics"),
+                                  tr("Allow OpenStudio Coalition to collect anonymous usage statistics to help improve the OpenStudio Application?"),
+                                  QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
```
jmarrec, line 336: * [ ] TODO: this is the last item needed to get this merged

jmarrec (changes requested):

src/openstudio_lib/AnalyticsHelper.cpp
macumber: `m_networkAccessManager` is parented by `this`

src/openstudio_lib/AnalyticsHelper.cpp
macumber: I think we'll have a hard enough time making sense of the tab data to start with.

src/openstudio_lib/MainWindow.cpp
macumber: Done

jmarrec (approved):

PR #566, opened 2022-10-13
 Enable C++20 and use nighty build of OpenStudio (3.5.0-alpha, E+ 22.2.0)

jmarrec:
Use the nightly build from the SDK as a lot of things already changed:

* conan dependencies have all been upgraded
* It uses C++20
* A few API breaks have been made (eg: MaterialAirWall is gone now)
* The CLI is no longer statically linked due to Python Measures being added

Should get us a head start and then only focus on 22.2.0 related problems


jmarrec:
**Note**: The build will be broken for now, as I found a bug, so I'm going to be using the installers from https://github.com/NREL/OpenStudio/pull/4712 once they post

CMakeLists.txt
```diff
@@ -740,15 +693,31 @@ endif()
...
 ######### Install SDK dependencies here ################
 
-#install(DIRECTORY "${openstudio_ROOT_DIR}/EnergyPlus" DESTINATION "." COMPONENT "OpenStudioApp" USE_SOURCE_PERMISSIONS)
+# The openstudio::openstudio target (which is the cli) includes a version string
+# in the file name. We don't want that degree of coupling between app and sdk
+# (we want the app to be able to use different sdk version without recompile,
+# so this project (the app) will install and identify the cli generically as
+# openstudio or openstudio.exe on windows. The os_cli_install_name is configured
+# into the utilities target so we need to set it before the uilities target is defined
+# TODO: THis isn't the case anymore
```
macumber, line 702: What isn't the case anymore?

ConanInstall.cmake
```diff
@@ -31,23 +31,36 @@ if(NOT CONAN_OPENSTUDIO_ALREADY_RUN)
...
-  conan_check(VERSION 1.43.0 REQUIRED)
+  conan_check(VERSION 1.48.0 REQUIRED)
 
   message(STATUS "openstudio: RUNNING CONAN")
 
   # Add NREL remote and place it first in line, since we vendored dependencies to NREL's repo, they will be picked first
   # TJC 2021-04-27 bintray.com is decommissioned as of 2021-05-01. See commercialbuildings as replacement below.
-  conan_add_remote(NAME nrel INDEX 0
-     URL https://conan.openstudio.net/artifactory/api/conan/openstudio)
+  if(LSB_RELEASE_ID_SHORT MATCHES "CentOS")
```
macumber, line 40: Why add support for CentOS?

src/shared_gui_components/OSGridController.hpp
```diff
@@ -110,9 +110,9 @@ class DataSource
   explicit DataSource(const std::function<std::vector<ItemType>(InputType)>& t_sourceFunc, bool t_wantsPlaceholder = false,
                       const QSharedPointer<BaseConcept>& t_dropZoneConcept = QSharedPointer<BaseConcept>())
     : m_wantsPlaceholder(t_wantsPlaceholder), m_dropZoneConcept(t_dropZoneConcept) {
-    typedef decltype(t_sourceFunc) IncommingFuncType;
-    typedef typename std::remove_reference<typename std::remove_cv<IncommingFuncType>::type>::type FunctionType;
-    typedef typename std::remove_reference<typename std::remove_cv<typename FunctionType::argument_type>::type>::type ParamType;
+
+    // Remove const and reference qualifiers: passing a function that takes `const ModelObject&` => `ModelObject` so we can use that for casting
+    using ParamType = typename std::remove_reference<typename std::remove_cv<InputType>::type>::type;
```
macumber, line 115: This is pretty wild, I didn't know this existed, but I guess it's been there since C++11.

macumber:
This is great @jmarrec, nice job!  I wonder if the https://github.com/NREL/openstudio-common-measures-gem/tree/develop/lib/measures/view_model and https://github.com/NREL/openstudio-common-measures-gem/tree/develop/lib/measures/view_data will need to be updated to handle the air wall material removal.  I don't see that referenced in there but will want to try it out.

src/shared_gui_components/BCLMeasureDialog.cpp
```diff
@@ -216,7 +216,8 @@ boost::optional<openstudio::BCLMeasure> BCLMeasureDialog::createMeasure() {
       result->changeUID();
 
       // change the files on disk
-      result->updateMeasureScript(m_bclMeasureToCopy->measureType(), measureType, m_bclMeasureToCopy->className(), className, name, description,
+      result->updateMeasureScript(m_bclMeasureToCopy->measureType(), measureType, m_bclMeasureToCopy->measureLanguage(),
```
macumber, line 219: Is NREL going to post documentation on how to use Python measures?  Can you mix Ruby and Python measures in a workflow?

src/openstudio_lib/ScriptFolderListView.cpp
```diff
@@ -88,7 +88,7 @@ void ScriptFolderListView::addScriptToFolder(const openstudio::path& t_path, con
   openstudio::filesystem::create_directories(folder);
   openstudio::path filename = folder / t_path.filename();
   filename = iterateFileName(filename);
-  openstudio::filesystem::copy_file(t_path, filename, openstudio::filesystem::copy_option::overwrite_if_exists);
```
macumber, line 91: I guess OS updated to std::filesystem instead of boost::filesystem?

src/shared_gui_components/OSGridController.hpp
jmarrec: The fix is simple and logical, but I had to test it out to make sure I wasn't doing something stupid (it felt off to see the complication given than InputType is passed as a template param to begin with...)

https://gcc.godbolt.org/z/6WdErGzbe

src/shared_gui_components/BCLMeasureDialog.cpp
jmarrec: Ultimately you will be able to mix them. This is experimental and for demonstration purposes right now (mixing measure types works on Mac and Windows, fails on Ubuntu, at the moment).
Aside from that they work exactly the same, syntax aside. Same layout, same methods. And the BCL now has a template for python measures too. 

Try this out in Ruby if you want to test it:

```
OpenStudio::BCLMeasure.new('A Test Python Measure', 'ATestPythonMeasure', 'atestpythonmeasure', 'enveloppe.fenestration', 'ModelMeasure'.to_MeasureType, 'Testing out the Python Measure', 'This is a measure that is written in Python', 'Python'.to_MeasureLanguage)
```

The installers also ship with new OSW in Examples/

src/openstudio_lib/ScriptFolderListView.cpp
jmarrec: no. But boost was updated to 1.79.

PR #573, opened 2022-10-24
Run Clang checks: `performance-for-range-copy` `readability-qualified-auto`, `modernize-use-auto`, `modernize-use-using -fix`

jmarrec:


macumber:
This all looks good to me, is the performance for range based loops check not smart enough to know when it can't use const?  I saw you had to manually fix that.

macumber (approved):
Looks good!

macumber:
Merge when you are ready!

jmarrec:
> This all looks good to me, is the performance for range based loops check not smart enough to know when it can't use const?  I saw you had to manually fix that.

It misfires sometimes yes. There are more clang tidy checks that could fix it further though, but I'm trying to do something targeted so not enabling 100 checks and ending up with something impossible to review and even spot check.

PR #595, opened 2023-03-31
Upgrade to OS SDK 3.6.0 alpha

jmarrec:
https://github.com/NREL/OpenStudio/commit/77ffac62c5cf2824e74bfe996f6c4a71adee9823


jmarrec:
the arm64 package sha is correct, but it's not on S3 yet.

src/shared_gui_components/OSGridController.hpp
```diff
@@ -276,6 +281,23 @@ class OSGridController : public QObject
...
+                          std::function<bool(DataSourceType*, const ValueType&)> setter,
+                          const boost::optional<std::function<void(DataSourceType*)>> reset = boost::none,
+                          const boost::optional<std::function<bool(DataSourceType*)>> isDefaulted = boost::none,
+                          const boost::optional<DataSource>& t_source = boost::none) {
+    m_baseConcepts.push_back(makeDataSourceAdapter(
+      QSharedPointer<ValueEditConcept<ValueType>>(new ValueEditConceptImpl<ValueType, DataSourceType>(heading, getter, setter, reset, isDefaulted)),
+      t_source));
+  }
+
+  // Corner case when you have a true optional with no default value
```
jmarrec, line 300: new overload for const T& setter. During the https://github.com/NREL/OpenStudio/pull/4532 I made all setters that take a `std::string` take instead a `const std::string&`

PR #599, opened 2023-04-12
Fix #145 - Allow reordering HVACComponent on Loop via drag and drop

### LINKED ISSUE #145: Allow rearraging of AirLoop and PlantLoop components
It would be nice to be able to rearrange the components (fans, pumps, coils, boilers, chillers, SetPointManagers, etc.) of an AirLoopHVAC or PlantLoop in the app by dragging and dropping an object rather than removing it and then adding a new object from the Library.

jmarrec:
* Fix #145

reorder_loop_comps

As can be seen above, I can't move a Node, or a splitter / mixer, or an OutdoorAirSystem

## TESTING WELCOME: TRY CREATIVE WAYS TO MAKE IT CRASH PLEASE!

### DIFF
--- a/src/openstudio_lib/GridItem.cpp
+++ b/src/openstudio_lib/GridItem.cpp
@@ -127,6 +127,8 @@
 #include <iterator>
 #include <utility>
 
+static constexpr bool EXTRA_DEBUG = false;
+
 using namespace openstudio::model;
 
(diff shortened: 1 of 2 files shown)


macumber:
This looks pretty awesome!  I will try it out and see if I can find any bugs.

Issue #606, opened 2023-05-15
Remove windows2019 runner from CI in favor of 2022


jmarrec:
Issue overview
--------------

We are in a situation where for a Release, both the 2019 and 2022 are getting signed and whichever finishes last ends up being the one uploaded to the release...

jmarrec:
Fixed via #607

PR #607, opened 2023-05-15
Send tab name instead of tab index and fix CORS issue

macumber:


src/openstudio_lib/OSWebEnginePage.cpp
```diff
@@ -42,8 +44,15 @@ OSUrlRequestInterceptor::OSUrlRequestInterceptor(QObject* parent) : QWebEngineUr
...
+  static QByteArray userAgent;
+  if (userAgent.isEmpty()) {
+    userAgent.append("OpenStudioApp/");
+    userAgent.append(OpenStudioApplicationVersionMajor());
+    userAgent.append(".");
+    userAgent.append(OpenStudioApplicationVersionMinor());
+  }
   info.setHttpHeader("Accept-Language", "en-US,en;q=0.9,es;q=0.8,de;q=0.7");
-  info.setHttpHeader("Access-Control-Allow-Origin", "*");
+  info.setHttpHeader("User-Agent", userAgent);
```
macumber, line 55: @jmarrec the openstreetmap website was mad at us for not setting a proper user agent per their usage guidelines: https://operations.osmfoundation.org/policies/nominatim/

Also, Access-Control-Allow-Origin is a response header and we were triggering CORS checking by including it, removing it means we don't care about CORS

src/openstudio_lib/AnalyticsHelper.cpp
```diff
@@ -56,7 +56,7 @@ AnalyticsHelper::~AnalyticsHelper() {
   delete m_networkAccessManager;
 }
 
-void AnalyticsHelper::sendAnalytics(const QString& analyticsId, int verticalTabIndex) {
+void AnalyticsHelper::sendAnalytics(const QString& analyticsId, const std::string& contentType, const std::string& contentId) {
```
macumber, line 59: Thought we could eventually use this from SketchUp Plug-in

src/openstudio_lib/OSWebEnginePage.cpp
jmarrec: Hpp:

```
private:
  QByteArray userAgent;
```

ctor:
```
OSUrlRequestInterceptor::OSUrlRequestInterceptor(QObject* parent)
  : QWebEngineUrlRequestInterceptor(parent),
    userAgent(
      QByteArray::fromStdString(fmt::format("OpenStudioApp/{}.{}", OpenStudioApplicationVersionMajor(), OpenStudioApplicationVersionMinor()))) {}
```

jmarrec (approved):

macumber:
You are right @jmarrec, member variable is better.  I was just worn out from investigating that weird error and went with the first thing that worked, I'll clean it up now.

macumber:
Fixes #606, didn't remove the runner just yet but do only sign one from 2022

PR #609, opened 2023-05-21
Don't crash if LocalBCL is corrupt

### LINKED ISSUE #608: Windows 10: Application fails on launch on new computer with "D:\OSN\src\utilities\blc\Local.cpp@74  Unable to update Local Database"
Issue overview
--------------

The application fails on launch in on new computer with "D:\OSN\src\utilities\blc\Local.cpp@74  Unable to update Local Database"
Windows 10

I have tried many versions.  I thought maybe the data base was loaded in a previous version but not the updated one

## Current Behavior

On open the following screen is shown, followed by the next.

image
image

### DEBUG log messages 
image
#### Gives:
openstudio.OSDocument] <-2> OSDocument::fixWeatherFileInTemp: 1
[openstudio.OSDocument] <-2> No weather file
[openstudio.LocalLibraryController] <-2> Creating LocalLibraryController with base app 00000099EC1FF590
[OSGridView] <-2> Creating cell wrapper row=0, col=0
[OSGridView] <-2> Creating cell wrapper row=0, col=1
[OSGridView] <-2> Creating cell wrapper row=0, col=2
[OSGridView] <-2> Creating cell wrapper row=0, col=3
[OSGridView] <-2> Creating cell wrapper row=0, col=4
[OSGridView] <-2> Creating cell wrapper row=0, col=5
[openstudio.bcl.LocalBCL] <1> Unable to prepare dbVersion 1.0/1.1 Statement
[openstudio.bcl.LocalBCL] <2> Unable to update Local Database
[OpenStudio] <2> An unhandled exception has occurred: D:\OSN\src\utilities\bcl\LocalBCL.cpp@74 : Unable to update Local Database
[removeDir] <-2> Removed 7 files

[osapp.log](https://github.com/openstudiocoalition/OpenStudioApplication/files/11503467/osapp.log)

## Expected Behavior

It should open the application!

## Steps to Reproduce

1. Turn Off the firewall and real-time scanning ( McAfee)
2. Download... https://github.com/openstudiocoalition/OpenStudioApplication/releases/download/v1.5.0/OpenStudioApplication-1.5.0+2a9ad1764a-Windows.exe  at
image

3.  Try both the command line and icon to start the application
4.  Never gets past the second info box!

### Environment

Processor	Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz   3.40 GHz
Installed RAM	12.0 GB (11.9 GB usable)
Product ID	xxxxxxxxxxxxxxxxxxxxxxxx
System type	64-bit operating system, x64-based processor
Pen and touch	Pen support

### Context

I can not even start this application.  I do have an old Openstudio 2.9.1 running well

macumber:
Fixes #608

### DIFF
--- a/src/openstudio_lib/OSDocument.cpp
+++ b/src/openstudio_lib/OSDocument.cpp
@@ -1380,7 +1380,7 @@ boost::optional<BCLMeasure> OSDocument::getLocalMeasure(const std::string& uid) 
   if (m_haveLocalBCL) {
     try {
       result = LocalBCL::instance().getMeasure(uid);
-    }catch(const std::exception& e) {
+    } catch (const std::exception& e) {
       LOG(Error, "Cannot access local BCL: " << e.what());
       m_haveLocalBCL = false;
     }
--- a/src/openstudio_lib/OSDocument.cpp
+++ b/src/openstudio_lib/OSDocument.cpp
@@ -1349,14 +1349,89 @@ void OSDocument::addStandardMeasures() {
   enable();
 }
 
-boost::optional<BCLMeasure> OSDocument::standardReportMeasure() {
-  std::string uid("a25386cd-60e4-46bc-8b11-c755f379d916");
-  boost::optional<BCLMeasure> result = LocalBCL::instance().getMeasure(uid);
-  if (result) {
-    return result;
+boost::optional<BCLComponent> OSDocument::getLocalComponent(const std::string& uid) const {
+  boost::optional<BCLComponent> result;
+  if (m_haveLocalBCL) {
+    try {
+      result = LocalBCL::instance().getComponent(uid);
+    } catch (const std::exception& e) {
+      LOG(Error, "Cannot access local BCL: " << e.what());
+      m_haveLocalBCL = false;
+    }
   }
+  return result;
+}
 
-  if (RemoteBCL::isOnline()) {
+boost::optional<BCLComponent> OSDocument::getLocalComponent(const std::string& uid, const std::string& versionId) const {
+  boost::optional<BCLComponent> result;
+  if (m_haveLocalBCL) {
+    try {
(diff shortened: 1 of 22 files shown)


src/openstudio_lib/OSDocument.cpp
```diff
@@ -1349,14 +1349,89 @@ void OSDocument::addStandardMeasures() {
...
+
+boost::optional<BCLComponent> OSDocument::getLocalComponent(const std::string& uid, const std::string& versionId) const {
+  boost::optional<BCLComponent> result;
+  if (m_haveLocalBCL) {
+    try {
+      result = LocalBCL::instance().getComponent(uid, versionId);
+    } catch (const std::exception& e) {
+      LOG(Error, "Cannot access local BCL: " << e.what());
+      m_haveLocalBCL = false;
+    }
```
jmarrec, line 1373: My only comment is whether we shouldn't just try to get the standardMeasure first and set m_haveLocalBCL once and for all to avoid having to do this check over and over in each method. That's not a huge deal though.

In fact, aren't we getting it in most cases already, unless we're just reloading? https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/openstudio_lib/OSDocument.cpp#L233-L235

jmarrec (commented):
LGTM, suggested a change, I'm fine if you just merge this as-is though

src/openstudio_lib/OSDocument.cpp
macumber: I guess it is pretty unlikely that `m_haveLocalBCL` would be true and then false in a single session, there is some overhead to the try catch but it's probably minimal.  I would rather just protect against any path in case we change the initialization order at some point.  Let's just merge as is, thanks for looking!

PR #618, opened 2023-08-09
Add search bar to OSCollapsibleItemList

macumber:
Adds a search bar to OSCollapsibleItemList used for most of the library tabs.  OSListView which is used by the ScriptsTab is considerably different, so I did not add a search feature there for now.

### DIFF
--- a/src/openstudio_lib/OSCollapsibleItemList.cpp
+++ b/src/openstudio_lib/OSCollapsibleItemList.cpp
@@ -36,6 +36,7 @@
 #include <openstudio/utilities/core/Assert.hpp>
 
 #include <QHBoxLayout>
+#include <QLineEdit>
 #include <QPainter>
 #include <QPushButton>
 #include <QResizeEvent>
@@ -50,6 +51,7 @@ OSCollapsibleItemList::OSCollapsibleItemList(bool addScrollArea, QWidget* parent
     m_vLayout(nullptr),
     m_contentLayout(nullptr),
     m_selectedCollapsibleItem(nullptr),
+    m_searchActive(false),
     m_itemsDraggable(false),
     m_itemsRemoveable(false),
     m_itemsType(OSItemType::ListItem) {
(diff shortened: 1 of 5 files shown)


jmarrec (commented):
Just some nitpicks. Otherwise this is great.

Also, I think these would be good improvements, either for here directly or filed as tickets for later.

* some kind of button to clear the text in the search bar would be helpful.
* Hiding the collapsible sections that didn't match at all would be make it clearer

jmarrec:
@macumber This is great! here's a Peek capture.

618_Search

macumber:
I skipped adding the search bar to OSItem because that is used to display all objects of a single type.  There are two instances of OSItem.  One is under each OSCollapsibleItemHeader in an OSCollapsibleItemList (box on the right) so a second search bar would be redundant. The other is in the ModelObjectListView used on tabs that don't have a grid view (box on the left).  We could configure that search for the case on the left but I don't think it's as useful there.

image

Another place OSCollapsibleItemList is used is the load definition tab, not sure if we'd want to disable the search bar here for consistency with ModelObjectListView?

image

Finally, the other place a search bar might be useful would be on the OSGridView.

src/openstudio_lib/OSCollapsibleItemList.cpp
```diff
@@ -212,14 +214,63 @@ void OSCollapsibleItemList::onItemSelected(OSItem* item) {
...
+        }
+        //collapsibleItem->itemList()->clearSelection();
+      }
+    }
+  }
+}
+
+void OSCollapsibleItemList::onSearchTextEdited(const QString& text) {
+  QLayoutItem* layoutItem = nullptr;
+  OSCollapsibleItem* collapsibleItem = nullptr;
```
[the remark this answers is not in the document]
macumber: Done!

src/openstudio_lib/OSCollapsibleItemList.cpp
```diff
@@ -62,6 +62,7 @@ OSCollapsibleItemList::OSCollapsibleItemList(bool addScrollArea, QWidget* parent
   this->setLayout(outerVLayout);
 
   m_searchBox = new QLineEdit();
+  m_searchBox->setClearButtonEnabled(true);
```
jmarrec, line 65: Lol, that icon is a broom?

jmarrec (approved):
Love the adjustments, thanks @macumber this is going to be very exciting for users! Here's a Peek, v2:

618_Search_v2

src/openstudio_lib/OSCollapsibleItemList.cpp
macumber: Hehe, weird on Windows it is a nice x button.

PR #622, opened 2023-08-22
Fix #560 - Refactor HVACSystems view to hide unavailable controls, allow zooming display on VRF + Refrigeration tab

### LINKED ISSUE #560: HVAC tab for VRF does not zoom
Issue overview
--------------

Zoom buttons do not function.  Also, unable to return to the VRF main overview page.
## Steps to Reproduce

1.
2.
3.
4.

### Environment

 - Platform (Operating system, version): Windows 11 Pro

jmarrec:
* Fix #560

* Allow zooming the **display** on the VRF and on the Refrigeration tabs
     * Zoom is reset when you switch to/from the system list view  from/to the detailed view
* Replace the buttons to expand a system / move back up to the systems list view (+ tooltip)
* Hide the "Control" and "Grid" buttons when they aren't available.
    * When you were on it and switch to a system type (VRF / SHW, Refrigeration) for which it isn't available, you see the tab that says it's not available, the button is there to make it clear you're on it. As soon as you switch to a valid tab, it disappears.
    

Here is a GIF showing the new UI:

560_zoom_VRF_Refrigeration


src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -358,37 +356,49 @@ void HVACSystemsController::update() {
         m_hvacSystemsView->mainViewSwitcher->setView(m_refrigerationGridController->refrigerationGridView());
       } else {
         // Not allowed
+        // TODO: should we automatically switch to the topologyView?
```
macumber, line 359: I don't know if it would be more confusing to switch automatically or not, but this change to make it clear is at least an improvement.

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -331,22 +326,25 @@ void HVACSystemsController::update() {
...
       if (m_hvacSystemsView->hvacToolbarView->topologyViewButton->isChecked()) {
         m_refrigerationController = std::make_shared<RefrigerationController>();
 
         m_hvacSystemsView->mainViewSwitcher->setView(m_refrigerationController->refrigerationView());
-      } else if (m_hvacSystemsView->hvacToolbarView->gridViewButton->isChecked()) {
-        // TODO
 
-        m_hvacSystemsView->hvacToolbarView->zoomInButton->hide();
-        m_hvacSystemsView->hvacToolbarView->zoomOutButton->hide();
+        connect(m_hvacSystemsView->hvacToolbarView->zoomInButton, &QPushButton::clicked, m_refrigerationController->refrigerationView(),
```
macumber, line 337: I'd be worried about the buttons being shown but not connected or connected to the wrong thing.  Would it be a good idea to start them out hidden and then you could see when they are shown the connections are right below?

jmarrec: We own the code, and we're unlikely going to change this often. I'd say if it works, it works no?

I'm not opposed to the change, it's just some amount of work (maybe couple of hours with annoying manual testing) to change the logic from being show-default to hide-default. I'm not worried enough about maintainability, but if you really are, I support it.

macumber: I'm really less familiar with this code so it's probably just my initial impression, no need to clean it up if its working

PR #623, opened 2023-08-22
Fix #238 - Add a search bar for the Output Variable list, allow displaying only selected, bulk setting reporting frequency

### LINKED ISSUE #238: Possible option of a search icon in the output variable tab
Enhancement Request
Having a search icon in output variable tab can facilitate the user to search the desired output variable one wants to look for.

In the OS app, one have to manually scroll and look for the output variable he/she wants to select which takes unnecessary time this feature could save a lot of time.

output

jmarrec:
* Fix #238 
* Add a search bar for the Output Variable list, allow displaying only selected, bulk setting reporting frequency

238_OutputVariable

### DIFF
--- a/src/openstudio_lib/VariablesTabView.cpp
+++ b/src/openstudio_lib/VariablesTabView.cpp
@@ -443,7 +443,7 @@ void VariablesList::updateVariableList() {
     }
   }
 
-  for (auto& itr : potentialOutputVariableMap) {
+  for (const auto& itr : potentialOutputVariableMap) {
     const PotentialOutputVariable& pov = itr.second;
 
     //LOG(Debug, "Creating VariableListItem for: " << pov.name << ", " << pov.keyValue);
--- a/src/openstudio_lib/VariablesTabView.cpp
+++ b/src/openstudio_lib/VariablesTabView.cpp
@@ -35,6 +35,7 @@
 #include <openstudio/model/OutputVariable_Impl.hpp>
 
 #include "../model_editor/Utilities.hpp"
+#include "../shared_gui_components/OSDialog.hpp"
 
 #include <openstudio/utilities/sql/SqlFileEnums.hpp>
 
@@ -129,6 +130,12 @@ void VariableListItem::setVariableEnabled(bool t_enabled) {
 
 bool VariableListItem::isVariableEnabled() const {
   return m_onOffButton->isChecked();
+}
+
+void VariableListItem::setReportingFrequency(const std::string& freq) {
+  if (m_variable) {
+    m_variable->setReportingFrequency(freq);
+  }
 }
 
 void VariableListItem::onOffClicked(bool t_on) {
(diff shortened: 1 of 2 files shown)


src/openstudio_lib/VariablesTabView.cpp
```diff
@@ -139,34 +166,89 @@ void VariableListItem::onOffClicked(bool t_on) {
...
-  m_allOffBtn = new QPushButton("All Off");
-  m_allOffBtn->setFlat(true);
-  m_allOffBtn->setObjectName("StandardGrayButton");
-  innerbox->addWidget(m_allOnBtn);
-  innerbox->addWidget(m_allOffBtn);
+    connect(m_allOnBtn, &QPushButton::clicked, this, &VariablesList::allOnClicked);
+    connect(m_allOffBtn, &QPushButton::clicked, this, &VariablesList::allOffClicked);
+  }
+
+  outerHBox->addLayout(allOnOffVBox);
```
macumber, line 200: I think my main comment on this PR would be to remove the separated frame, then make the "All On" and "All Off" buttons apply to the currently visible output variables.  That would make it easier to turn on/off all variables for "Surfaces" for example.  The current implementation of turning on/off all variables in the model is not useful.

macumber: Here's an idea for the layout, I'd be happy to help on this if you'd like or you can go for it.

image

macumber: Any idea how up to date OpenStudio has been keeping the `ModelObject::outputVariableNames` methods?  I wonder if it would be better to use the RDD instead if it exists?

jmarrec: I would reckon some output variable are outdated as changes slipped through the cracks, but we do try to keep em up to date as part of the update to new E+ IO Freeze builds

src/openstudio_lib/VariablesTabView.cpp
```diff
@@ -183,8 +265,45 @@ void VariablesList::allOffClicked() {
...
 
 void VariablesList::enableAll(bool t_enabled) {
-  for (auto itr = m_variables.begin(); itr != m_variables.end(); ++itr) {
-    (*itr)->setVariableEnabled(t_enabled);
+  for (auto& m_variable : m_variables) {
+    if (m_variable->isVisible()) {
+      m_variable->setVariableEnabled(t_enabled);
+    }
+  }
+}
```
jmarrec, line 273: @macumber Here is where the "All On" button only affects Visible variables already.

src/openstudio_lib/VariablesTabView.cpp
jmarrec: The "All On" / "All Off" only applies to currently visible variables already. You can also see it in action in the GIF above.

If you think your layout is clearer I'm not opposed to it.

jmarrec:
Merging. Opened https://github.com/openstudiocoalition/OpenStudioApplication/issues/626 for follow up.

src/openstudio_lib/VariablesTabView.cpp
macumber: Oh cool, ok thanks

src/openstudio_lib/VariablesTabView.cpp
macumber: Moved to https://github.com/openstudiocoalition/OpenStudioApplication/issues/626

PR #624, opened 2023-08-26
Fix MacOS CI

macumber:
Remove package config files for OpenSSLv3


macumber:
@jmarrec FYI, I tested the output package and everything looks good

PR #628, opened 2023-08-28
Geometry Preview refactor - separate into several files, allow running in browser directly

jmarrec:
I find the geometry_preview.html awful to work with, as it has a lot of embedded jS into it.

Also, having the ability to run it directly into your browser is helpful for testing out changes.

This PR:
* Extract the inline jS from geometry_preview.html into separate jS files
* Upgrades these dependencies (to a version without major breaking changes), and downloads the unminified version that matches, so one could technically swap to these to debug something
* Be explicit about which versions of these dependencies we use
* Allows running the geometry_preview in your browser directly with some json data.
    * Add separate entry points `runFromJSON` and `runFromFile`


macumber:
This looks great @jmarrec, as an FYI, I've tried to keep the html and code here in sync with:

- https://github.com/NREL/openstudio-common-measures-gem/blob/develop/lib/measures/view_model/resources/report.html.in
- https://github.com/NREL/openstudio-common-measures-gem/blob/develop/lib/measures/view_data/resources/report.html.in

NREL uses those measures to view models from PAT, etc.  If NREL is willing to support you doing it could you consider porting these changes back to the `view_model` and `view_data` measures?

jmarrec:
Backported at https://github.com/NREL/openstudio-common-measures-gem/pull/156

Tested the measures locally + inside the OS App (I ended up embedding the jS files + data into the report.html to avoid issues)

PR #629, opened 2023-08-28
New Geometry Diagnostics in the Geometry Preview: Detect incorrectly oriented surfaces in spaces, non convex spaces

jmarrec:
Opt-in feature to enable the geometry diagnostics from this PR:

* https://github.com/NREL/OpenStudio/pull/4843

Convexity

# TODO:

* Having trouble tweaking the display of the dat GUI menu so that the labels won't be truncated as much.
    * The stylesheet is inlined in dat GUI. This can be seen here in the geometry_refactor branch: https://github.com/openstudiocoalition/OpenStudioApplication/blob/4d89d8bcba5bac67c519b00eb75aad76f6ac743a/src/openstudio_lib/library/js/dat.gui.0.7.9.js#L1673
* Not sure if I should build upon #628 or not. So I have two separate branches (one with 628, one without (this one))   
    *  if #628 is merged first, then should change the feature branch to `geometry_diagnostics_with_refactor`, see  https://github.com/openstudiocoalition/OpenStudioApplication/compare/geometry_refactor...geometry_diagnostics_with_refactor?expand=1

### DIFF
--- a/src/openstudio_lib/GeometryPreviewView.cpp
+++ b/src/openstudio_lib/GeometryPreviewView.cpp
@@ -30,6 +30,7 @@
 #include "GeometryPreviewView.hpp"
 #include "OSAppBase.hpp"
 #include "OSDocument.hpp"
+#include "MainWindow.hpp"
 
 #include "../model_editor/Application.hpp"
 
@@ -46,6 +47,7 @@
 #include <QFile>
 #include <QWebEngineScriptCollection>
 #include <QtConcurrent>
+#include <QCheckBox>
 
 using namespace std::placeholders;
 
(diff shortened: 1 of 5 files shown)


macumber:
Does this really need it's own separate html page?  I thought we just added these options to the main geometry view?

src/openstudio_lib/GeometryPreviewView.cpp
```diff
@@ -156,9 +180,13 @@ void PreviewWebView::onLoadFinished(bool ok) {
     std::function<void(double)> updatePercentage = std::bind(&PreviewWebView::onTranslateProgress, this, _1);
     //ThreeScene scene = modelToThreeJS(m_model.clone(true).cast<model::Model>(), true, updatePercentage); // triangulated
 
+    // qDebug() << "ThreeJSForwardTranslator";
+
     model::ThreeJSForwardTranslator ft;
-    ThreeScene scene = ft.modelToThreeJS(m_model, true, updatePercentage);  // triangulated
-    std::string json = scene.toJSON(false);                                 // no pretty print
+    m_includeGeometryDiagnostics = m_geometryDiagnosticsBox->isChecked();
+    ft.setIncludeGeometryDiagnostics(m_includeGeometryDiagnostics);
```
macumber, line 187: Is there a downside to just always including the geometry diagnostics on this page?  Is it significantly slower?

jmarrec:
> Does this really need its own separate html page? I thought we just added these options to the main geometry view?

I don't get the question. This is added to the main geometry view already (?!)

src/openstudio_lib/GeometryPreviewView.cpp
jmarrec: After I implemented caching, it's not that bad, but yes, it's slightly slower. See benchmarks here: https://github.com/NREL/OpenStudio/pull/4843#discussion_r1156098275

With caching:

| Test case                         | Release | Release   | Debug   | Debug     |
|-----------------------------------|---------|-----------|---------|-----------|
|                                   | No Diag | With Diag | No Diag | With Diag |
| RefBldgOutPatientNew2004_Chicago  | 1.39    | 1.4       | 10.09   | 10.19     |
| 7_7_Windows_Complete              | 3.79    | 3.95      | 27.17   | 28.33     |

jmarrec:
@macumber Closing this PR in favor of #630 and moving the comments/conversation over there.

PR #630, opened 2023-08-29
New Geometry Diagnostics in the Geometry Preview: Detect incorrectly oriented surfaces in spaces, non convex spaces

jmarrec:
* Supersedes #629 
* because it includes #628

----

## Opt-in feature to enable the geometry diagnostics from this PR:

* https://github.com/NREL/OpenStudio/pull/4843

Convexity

# TODO:

* Having trouble tweaking the display of the dat GUI menu so that the labels won't be truncated as much.
    * The stylesheet is inlined in dat GUI. This can be seen here: https://github.com/openstudiocoalition/OpenStudioApplication/blob/6aee38e6f54b8b7a851ec5054f94c6b3aab71a47/src/openstudio_lib/library/js/dat.gui.0.7.9.js#L1673

### DIFF
--- a/src/openstudio_lib/GeometryPreviewView.cpp
+++ b/src/openstudio_lib/GeometryPreviewView.cpp
@@ -106,7 +106,8 @@ PreviewWebView::PreviewWebView(bool isIP, const model::Model& model, QWidget* t_
   m_geometryDiagnosticsBox = new QCheckBox();
   m_geometryDiagnosticsBox->setText("Geometry Diagnostics");
   m_geometryDiagnosticsBox->setChecked(verboseOutput);
-  m_geometryDiagnosticsBox->setToolTip("Enables checks for Surface/Space Convexity. The ThreeJS export is slightly slower");
+  m_geometryDiagnosticsBox->setToolTip(
+    "Enables adjacency issues. Enables checks for Surface/Space Convexity, due to this the ThreeJS export is slightly slower");
   connect(m_geometryDiagnosticsBox, &QCheckBox::clicked, mainWindow, &MainWindow::toggleGeometryDiagnostics);
   connect(m_geometryDiagnosticsBox, &QCheckBox::stateChanged, [this](int state) {
     if (state == Qt::Checked && !m_includeGeometryDiagnostics) {
--- a/src/openstudio_lib/GeometryPreviewView.cpp
+++ b/src/openstudio_lib/GeometryPreviewView.cpp
@@ -30,6 +30,7 @@
 #include "GeometryPreviewView.hpp"
 #include "OSAppBase.hpp"
 #include "OSDocument.hpp"
+#include "MainWindow.hpp"
 
 #include "../model_editor/Application.hpp"
 
@@ -46,6 +47,7 @@
 #include <QFile>
 #include <QWebEngineScriptCollection>
 #include <QtConcurrent>
+#include <QCheckBox>
 
 using namespace std::placeholders;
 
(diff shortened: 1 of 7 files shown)


src/openstudio_lib/GeometryPreviewView.cpp
```diff
@@ -156,7 +180,11 @@ void PreviewWebView::onLoadFinished(bool ok) {
     std::function<void(double)> updatePercentage = std::bind(&PreviewWebView::onTranslateProgress, this, _1);
     //ThreeScene scene = modelToThreeJS(m_model.clone(true).cast<model::Model>(), true, updatePercentage); // triangulated
 
+    // qDebug() << "ThreeJSForwardTranslator";
+
     model::ThreeJSForwardTranslator ft;
+    m_includeGeometryDiagnostics = m_geometryDiagnosticsBox->isChecked();
+    ft.setIncludeGeometryDiagnostics(m_includeGeometryDiagnostics);
```
jmarrec, line 187: See original comments https://github.com/openstudiocoalition/OpenStudioApplication/pull/629#discussion_r1308156144

@macumber said:



> Is there a downside to just always including the geometry diagnostics on this page? Is it significantly slower?

jmarrec: After I implemented caching, it's not that bad, but yes, it's slightly slower. See benchmarks here: https://github.com/NREL/OpenStudio/pull/4843#discussion_r1156098275

With caching:

| Test case                         | Release | Release   | Debug   | Debug     |
|-----------------------------------|---------|-----------|---------|-----------|
|                                   | No Diag | With Diag | No Diag | With Diag |
| RefBldgOutPatientNew2004_Chicago  | 1.39    | 1.4       | 10.09   | 10.19     |
| 7_7_Windows_Complete              | 3.79    | 3.95      | 27.17   | 28.33     |

jmarrec:
https://github.com/openstudiocoalition/OpenStudioApplication/pull/629#issuecomment-1696715056

@macumber said:

> Does this really need its own separate html page? I thought we just added these options to the main geometry view?

I don't get the question. This is added to the main geometry view already (?!)

macumber:
> [](https://github.com/openstudiocoalition/OpenStudioApplication/pull/629#issuecomment-1696715056)
> 
> @macumber said:
> 
> > Does this really need its own separate html page? I thought we just added these options to the main geometry view?
> 
> I don't get the question. This is added to the main geometry view already (?!)

Sorry, I meant "does this need its own checkbox and page reload"?  Based on your timing stats it looks like the penalty for diagnostics is pretty slow, what do you think about just including the diagnostics all the time?

jmarrec:
The page reload is only to change the datGUI menu. And if the case you go from no diag to diag, the JSON has to be generated again.

I don't know. I'm not opposed to having it enabled all the time if we think that's better and the penalty is irrelevant... 

I don't dislike the idea of being able to turn it off if you want though, less clutter in the menu, plus some users might have terribly complicated models where the penalty is actually significant (we've seem reports here in the issues where people have 1500 zones models...). The drawback is that you have a checkbox... but if that's the issue, perhaps we can hide that in the "Preferences" menu instead.

macumber:
I'll build it and try it out, either way is probably good.  The checkbox is nice in that it advertises the feature....

macumber:
I tried this out today and it seems to work correctly.  The only thing that confused me is that the "Diagnostic Tools" menu item is there regardless of the "Geometry Diagnostics" checkbox.  It seems that the checkbox should control that entire menu item, right now it just controls the "Surface Level Issues" and "Space Level Issues" items.  I know that is how it was before, but do you think that adding a "Geometry Diagnostics" checkbox should now control whether the "Diagnostic Tools" menu item is shown?

src/openstudio_lib/library/geometry_preview.html
```diff
@@ -1412,22 +1443,58 @@
...
-  var f2 = gui.addFolder('Potential Adjacency Issues');
-  f2.add(settings, 'findAdjacencyIssues').name('Enable').onChange(value => {
-      removeSelection();
-      setLocalStorage('findAdjacencyThreshold', value);
-      updateAdjacencyIssues();
-      update(value);
-  });
-  f2.add(settings, 'findAdjacencyThreshold', 0.1, 10, 0.5).name('Threshold').onFinishChange(function (value) {
-      if (settings.findAdjacencyIssues) {
+  var diagnostics = gui.addFolder("Diagnostic Tools")
```
macumber, line 1446: Consider moving all "Diagnostic Tools" under check for `includeGeometryDiagnostics`?

jmarrec: So, the geometry diagnostics checkbox affect the way the ThreeJS translation is done (include those, with computational cost, or do not). The adjacency issue and co were already there and aren't affected.

If we think it's better to hide the entire menu we can do that though.

macumber: My only concern is that people might click "Geometry Diagnostics" and then not be able to find the new menu items.  Another idea would be to rename the checkbox "Extra Geometry Diagnostics" so people know that new items will be added under "Diagnostic Tools".  It's fine just how it is, only make any changes if you think it's an improvement.

macumber (approved):

src/openstudio_lib/library/geometry_preview.html
jmarrec: Hiding the entire "Diagnostics Tools" folder when the checkbox is unclicked. adbd297e6cdefc97e3f21633b1fb9c669a16d2d7

macumber: Thanks Julien, I think this is nicer

PR #631, opened 2023-09-06
Update variables layout

### LINKED ISSUE #626: Redo the layout of the new Output:Variable filter capability
Enhancement Request
--------------

See @macumber's comment here: https://github.com/openstudiocoalition/OpenStudioApplication/pull/623#discussion_r1306553255

> Here's an idea for the layout, I'd be happy to help on this if you'd like or you can go for it.

image

## Detailed Description

Determine if the proposed layout is clearer and implement it if so.

macumber:
Fixes #626

I just moved things around a little bit, I also added a progress bar when editing all the variables. 
image

### DIFF
--- a/.github/workflows/app_build.yml
+++ b/.github/workflows/app_build.yml
@@ -92,7 +92,7 @@ jobs:
           PLATFORM_NAME: Darwin
           CPACK_BINARY_DEB: OFF
           CPACK_BINARY_IFW: ON
-          CPACK_BINARY_TGZ: ON
+          CPACK_BINARY_TGZ: OFF
           CPACK_BINARY_ZIP: OFF
           BINARY_EXT: dmg
           COMPRESSED_EXT: tar.gz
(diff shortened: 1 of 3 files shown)


jmarrec:
Thanks @macumber. I'm fine with the changes, except the combobox for picking "", "Enabled", "Disabled".

I'd rather have several buttons.

Maybe "View: [All] [Enabled] [Disabled]", only one button is pressed at the same time, same a the ["Layout", "Grid", "Control"] buttons on the HVAC tab.

Either on its own line, or with some kind of separator from the Filter box.


Also, the progress bar is useless at least on my machine, evenhaving enabled all variables and then switching the frequency, the only thing I see on the display is the mere flicker of a progressbar.

macumber:
Wow @jmarrec so many options!  I like Atlernative4, I'd just get rid of the big blank space between the buttons and the vertical separatr. We can remove the progress bar for bulk updating frequency if you want, it's most useful for bulk turning variables on (which really just makes a bunch of model objects).

macumber:
Looks great!

jmarrec:
The macos-12 failed because of cpack, I haven't seen that before. https://github.com/openstudiocoalition/OpenStudioApplication/actions/runs/6132515635/job/16643468685?pr=631#step:19:1001

macumber:
GitHub updates these runner images all the time, I wonder if something changed to break this, they have release notes at https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md.  It looks like it affects the tar.gz package only so I'll try disabling that for MacOS for now

PR #639, opened 2023-11-08
Fix #638 - Make links clickable for measure descriptions

### LINKED ISSUE #638: Make links clickable in measure descriptions
Enhancement Request
--------------
Any `href` links I've used in measure descriptions render as links, but are not clickable.

## Detailed Description
Links currently can only be right-clicked -> "Copy Link Location".

## Possible Implementation
It's a QLabel that's used for formatting the name, units, and description, and by default this property is false: https://doc.qt.io/qt-6/qlabel.html#openExternalLinks-prop

jmarrec:
* Fix #638 
* Can't think of a side effect making this. We're not worried about phishing attempts

### DIFF
--- a/src/shared_gui_components/EditView.cpp
+++ b/src/shared_gui_components/EditView.cpp
@@ -179,6 +179,7 @@ DoubleInputView::DoubleInputView() : InputView() {
   setLayout(vLayout);
 
   nameLabel = new QLabel();
+  nameLabel->setOpenExternalLinks(true);
   nameLabel->setTextFormat(Qt::RichText);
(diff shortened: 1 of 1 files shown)


macumber:
Looks like this will open links in the system browser?  That seems fine, the measure writer could include malicious links, that would have to be taken care of by some sort of review process.  Give that, this seems ok to me

jmarrec:
I thought about this, i just think doing phishing via an OpenStudio Measure hosted on BCL would be... surprising :)

macumber (approved):

macumber:
Agreed :-)  Merge away!

PR #643, opened 2023-11-23
Fix bug when cloning a story with windows

macumber:
Includes build output of https://github.com/openstudiocoalition/floorspace.js/pull/1

Notice I had to disable the Uglify plugin in the build process here:

https://github.com/openstudiocoalition/floorspace.js/pull/1/files#diff-428f295cce6ac12808d1771cf1dcada74b6dd5c38ac92555ac07e6a2fc28a8b4R34

I was hoping NREL might be able to update the floorspace.js build scripts but sounds like that is not in their current plan


jmarrec:
@macumber I can't even see the diff because it's too large.

1. Could we separate the jS like I did on #628 ?
2. Can we figure out a way to uglify it anyways, even if that means manually editing the bit that is faulty?

macumber:
I was hoping NREL would be able to fix the build scripts, but now that we know that is not happening in the near future, I can try getting uglify to work again.

macumber:
@jmarrec I'm trying to look into this again, I put out a request on Fiverr to see if someone can update webpack from version 1 to version 4 or 5, I think that is what is needed to get the uglify working again (now it is just an option in webpack instead of a plugin). 

I don't quite understand your request about breaking the JavaScript out.  You want to separate the html from the JavaScript?  There is a `build-embeddable.js` script that makes the `embeddable_geometry_editor.html` (adds API for the OS App to use) and `standalone_geometry_editor.html` (can be opened in a local browser) files.  There really isn't much interesting html, it's almost all JavaScript.

jmarrec:
@macumber Yeah I was just wondering if we could have the HTML reference a .js file, so we could see the HTML structure a bit better. And we could potentially just locally change the minified with the full one during testing/debugging when needed. Something like

```diff
-<script src="js/floorspace.min.js"></script>
+<script src="js/floorspace.js"></script>
```

That'd allow to differentiate between external libs and our own code/wrappers around these libs.

I've found that helpful for the geometry_preview.html. But I'm not forcing anything here.

jmarrec:
@macumber In the interim, I'm perfectly happy if you want to merge this one and we open a subsequent one later

macumber:
I've updated this PR with the new minified version of FloorspaceJS built from https://github.com/openstudiocoalition/floorspace.js

macumber:
The html for the embedded editor that is separate from the built JavaScript is just hard coded at: https://github.com/openstudiocoalition/floorspace.js/blob/develop/build/build-embeddable.js#L19

src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1288,7 +1288,7 @@ void OpenStudioApp::startMeasureManagerProcess() {
   tcpServer.close();
 
   QString portString = QString::number(port);
-  QString urlString = "http://127.0.0.1:" + portString;
+  QString urlString = "http://localhost:" + portString;
```
macumber, line 1291: @jmarrec what do you think about this?  For some reason when I run the app from the installer it can't find the measure manager.  Is there some reason the new measure manager is only listening to localhost and not 127.0.0.1?  I always thought the two were the same for IPv4.

```
macumber@Automatic-Magic-PC MINGW64 /c/openstudioapplication-1.7.0-rc0/bin
$ ./openstudio classic measure -s 1616
[2023-12-18 20:53:58] INFO  WEBrick 1.6.0
[2023-12-18 20:53:58] INFO  ruby 2.7.2 (2020-10-01) [x64-mswin64_140]
[2023-12-18 20:53:58] INFO  WEBrick::HTTPServer#start: pid=17592 port=1616
127.0.0.1 - - [18/Dec/2023:20:54:15 Mountain Standard Time] "GET / HTTP/1.1" 200 79
- -> /
[2023-12-18 20:54:35] INFO  going to shutdown ...
[2023-12-18 20:54:35] INFO  WEBrick::HTTPServer#start done.
┌────────────────────────────────────────────────────────────────────────────────┐
│  The `classic` command is deprecated and will be removed in a future release   │
└────────────────────────────────────────────────────────────────────────────────┘

macumber@Automatic-Magic-PC MINGW64 /c/openstudioapplication-1.7.0-rc0/bin
$ ./openstudio measure -s 1616
MeasureManager ReadyAccepting requests on: http://localhost:1616/

```

macumber:
@jmarrec I'm planning to merge this and make an RC1 tomorrow, sound good?  I was going to try upgrading to Qt 6.6 as well but there is a bug in moc on Mac

src/shared_gui_components/MeasureManager.cpp
```diff
@@ -579,6 +579,10 @@ boost::optional<measure::OSArgument> MeasureManager::getArgument(const measure::
     Json::Value choiceValues = argument.get("choice_values", Json::Value(Json::arrayValue));
     Json::Value choiceDisplayNames = argument.get("choice_display_names", Json::Value(Json::arrayValue));
 
+    if (choiceValues.empty()) {
+      choiceValues = argument.get("choices_values", Json::Value(Json::arrayValue));
```
macumber, line 583: This is a bug in the OS repo, there is a typo of `choices_values` instead of `choice_values`

jmarrec: I fixed it already in https://github.com/NREL/OpenStudio/pull/5046 but it wasn't selected for 3.7.0 official...

src/openstudio_app/OpenStudioApp.cpp
jmarrec: You're on windows I assume? I kinda just gave up trying to make both localhost and 127.0.0.0 work on windows... It's a limitation of the cpprestsdk afaik. https://github.com/NREL/OpenStudio/blob/05c60768c1e1057083ce644d49711f7377f8ce1e/src/cli/MeasureManager.cpp#L563-L569

jmarrec:
> @jmarrec I'm planning to merge this and make an RC1 tomorrow, sound good? I was going to try upgrading to Qt 6.6 as well but there is a bug in moc on Mac

Sounds good. I'll go check #650

jmarrec:
Actually I just realized I never made a PR for the Python measures branch I had going on...

PR #647, opened 2023-12-18
Use the GitHub hosted macos-13 runners

macumber:
I thought that the macos-13-arm64 runner was available for free use, but it appears that it is not: https://github.com/actions/runner-images/issues/8439

Probably want to hold off on this until it is available


jmarrec:
Yeah they're paid. But maybe we upgrade our self hosted one?

macumber:
The Mac Mini is on MacOS 13.5 right now, I was just trying to see if I could get out of maintaining it...

PR #650, opened 2023-12-23
Update to Qt 6.6.1

macumber:


macumber:
Looks like this is hitting this on Mac: https://bugreports.qt.io/browse/QTBUG-117765

jmarrec:
Quoting the QT-BUG commments:

> I got this moc failure when using Xcode 14.2. However, with Xcode 14.3.1 it's fine. Both are on the same codebase.

Seems like our CI machine is at 14.2.

jmarrec:
Built just fine on my M1 mac, after fixing a build error due to boost, with Apple-clang 15.0.0

```
$ sw_vers
ProductName:		macOS
ProductVersion:		14.2.1
BuildVersion:		23C71
$ clang --version
Apple clang version 15.0.0 (clang-1500.1.0.2.5)
Target: arm64-apple-darwin23.2.0
Thread model: posix
```

jmarrec:
@macumber Should we just update xcode on the mac mini and roll with it?

macumber:
The Mac Mini was fine, it was the GitHub hosted Mac that had the problem.  I switched https://github.com/openstudiocoalition/OpenStudioApplication/pull/647/ to only update the GitHub hosted Mac to MacOS 13 and merged that in here.  Maybe that will work?

PR #651, opened 2023-12-23
Add support for Python Measures

jmarrec:


src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1304,14 +1321,18 @@ void OpenStudioApp::startMeasureManagerProcess() {
   test = (connect(m_measureManagerProcess, &QProcess::stateChanged, this, &OpenStudioApp::measureManagerProcessStateChanged) != nullptr);
   OS_ASSERT(test);
 
-  QString program = toQString(openstudioCLIPath());
+  const QString program = toQString(openstudioCLIPath());
   QStringList arguments;
+
+  if (m_useLabsCLI) {
+    arguments << "labs";
+  }
```
jmarrec, line 1329:
```suggestion
  if (!m_useLabsCLI) {
    arguments << "classic";
  }
```

src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1281,6 +1282,22 @@ void OpenStudioApp::startMeasureManagerProcess() {
...
 
+  // Debugging: attach to your own measure manager you've launched in a debugger prior to firing the OSApp.
+  if (qEnvironmentVariableIsSet("OPENSTUDIO_APPLICATION_USE_LOCAL_MEASURE_MANAGER_PORT")) {
+    LOG(Debug, "OPENSTUDIO_APPLICATION_USE_LOCAL_MEASURE_MANAGER_PORT is set");
+    bool ok = false;
+    const int port = qEnvironmentVariableIntValue("OPENSTUDIO_APPLICATION_USE_LOCAL_MEASURE_MANAGER_PORT", &ok);
+    if (ok) {
+      LOG(Debug, "OPENSTUDIO_APPLICATION_USE_LOCAL_MEASURE_MANAGER_PORT is " << port);
+      QString portString = QString::number(port);
+      QString urlString = "http://127.0.0.1:" + portString;
```
jmarrec, line 1293:
```suggestion
      QString urlString = "http://localhost:" + portString;
```

src/openstudio_lib/ApplyMeasureNowDialog.cpp
```diff
@@ -389,6 +390,9 @@ void ApplyMeasureNowDialog::runMeasure() {
   OS_ASSERT(tempWorkflowJSONPath);
 
   QStringList arguments;
+  if (OSAppBase::instance()->currentDocument()->mainWindow()->useLabsCLI()) {
+    arguments << "labs";
+  }
```
jmarrec, line 395:
```suggestion
  if (!OSAppBase::instance()->currentDocument()->mainWindow()->useLabsCLI()) {
    arguments << "classic";
  }
```

jmarrec (commented):
I'm rebasing this older branch, so I don't know if everything is good or not, but I found this adjustments I need to make due to the fact that the 3.7.0 installer has the C++ CLI as default now

src/openstudio_app/OpenStudioApp.cpp
```diff
@@ -1158,7 +1158,8 @@ void OpenStudioApp::readSettings() {
   setLastPath(settings.value("lastPath", QDir::homePath()).toString());
   setDviewPath(openstudio::toPath(settings.value("dviewPath", "").toString()));
   m_currLang = settings.value("language", "en").toString();
-  LOG_FREE(Debug, "OpenStudioApp", "\n\n\nm_currLang=[" << m_currLang.toStdString() << "]\n\n\n");
+  m_useLabsCLI = settings.value("useLabsCLI", false).toBool();
```
macumber, line 1161: Probably want to default to using the new CLI and provide option to use the old one?  For measure manager I just assumed we'd use the new one, do we really need to support the old CLI for measure manager or run purposes?

src/shared_gui_components/WorkflowController.hpp
```diff
@@ -137,6 +137,7 @@ class MeasureStepController : public WorkflowStepController
   int count() override;
 
   MeasureType measureType() const;
+  MeasureLanguage measureLanguage() const;
```
macumber, line 140: Woah, is this a feature now?

macumber:
Hey @jmarrec this is cool, my two main questions are:

1. Do you want to squeeze this in to 1.7.0?  If so, we should put it in the RC to get testing

macumber:
Hey @jmarrec this is great, my two main questions are:

1. I think currently develop is using the new CLI for everything right now.  Adding options to support the classic CLI will be more work and would have to be removed in a release or two anyway since the classic CLI won't be around forever.  My suggestion would be to remove all the options in this PR and just use the new CLI.
2. Do you want to squeeze this into 1.7.0?  If so we should put it into the first RC to get testing on it.

jmarrec:
2. Yes.
1. Because everything is still new, i definitely want an option to fall back on the classic CLI. We don't want people to not be able to run their workflows. It's trivial to remove in the future. Basically revert this PR

src/openstudio_app/OpenStudioApp.cpp
jmarrec: I'm not sure. Perhaps yes, we enable the default of OS SDK.
But yeah, we probably want the fallback to classic for both MeasureManager and run.

src/shared_gui_components/WorkflowController.hpp
jmarrec: Yes. Has to be. You need to know if it's a ruby or a python measure! (and the measure manager / workflow run command does need to know too)

Issue #659, opened 2023-12-31
Mac app locks up when opening file that requires update and weather file reset


macumber:
Issue overview
--------------
Reported by @brgix in detail here https://github.com/openstudiocoalition/OpenStudioApplication/issues/653#issuecomment-1872964638

### Environment

Some additional details about your environment for this issue (if relevant):
 - Platform (Operating system, version): Mac M1 13.5
 - Version of OpenStudioApplication (if using an intermediate build, include SHA): 1.7.0-rc1

jmarrec:
Testing with https://github.com/rd2/tbd/blob/402f9f5acaa95e85bfdae33d28ea115e125db683/spec/files/osms/in/warehouse.osm#L6865-L6877 

On Ubuntu, no issue. But I can reproduce the lockup of the App on Mac M1.

jmarrec:
Seems like the issue is the two QMessageBox atop each other or something. If I comment out the lines here: https://github.com/openstudiocoalition/OpenStudioApplication/blob/a1f218a2d16bbfad2c00ac11cc1d5e46f212a8e8/src/openstudio_lib/OSDocument.cpp#L356-L357

then it works fine.

jmarrec:
Bad workaround for now at #663

macumber:
Downgraded to regular bug since crash is fixed

macumber:
New scope of this bug would be to re-enable the warning to the user on Mac

PR #663, opened 2024-01-08
Bad workaround for #659

jmarrec:
Temporary workaround for #659 on mac. Just disabling the messagebox `Invalid weather file object, weather file object has been reset. Please choose another weather file`. Hopefully we find an actual fix, but in the meantime it's more important to be able to load a file than to see this warning messagebox.

### DIFF
--- a/src/openstudio_lib/OSDocument.cpp
+++ b/src/openstudio_lib/OSDocument.cpp
@@ -126,6 +126,7 @@
 #include <QIcon>
 #include <QInputDialog>
 #include <QSettings>
+#include <QtGlobal>  // Workaround for #659
 
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


macumber (approved):

macumber:
The MessageBox in OpenStudioApp::versionUpdateMessageBox is not parented, the one in OSDocument::weatherFileReset is, I wonder if that makes any difference?

Went looking for any related Qt bugs:

https://bugreports.qt.io/browse/QTBUG-116622
https://bugreports.qt.io/browse/QTBUG-114546

jmarrec:
@macumber I tried this, doesn't work.

```
--- a/src/openstudio_app/OpenStudioApp.cpp
+++ b/src/openstudio_app/OpenStudioApp.cpp
@@ -1076,7 +1076,13 @@ void OpenStudioApp::childEvent(QChildEvent* event) {

 void OpenStudioApp::versionUpdateMessageBox(const osversion::VersionTranslator& translator, bool successful, const QString& fileName,
                                             const openstudio::path& tempModelDir) {
-  QMessageBox messageBox;
+
+  QWidget* parent = nullptr;
+
+  if (this->currentDocument()) {
+    parent = this->currentDocument()->mainWindow();
+  }
+  QMessageBox messageBox(parent);
```

I also tried a few more things including OSAppBase::mainWidget as I said here: https://github.com/openstudiocoalition/OpenStudioApplication/issues/659#issuecomment-1880855544

PR #666, opened 2024-01-10
Test SDK CLI from installers from another C++ CI run

jmarrec:
manual dispatch, put run id for the one that produced the installers, and it tests it


macumber:
Amazing!  I never thought we had this many files in our installer...

PR #667, opened 2024-01-11
Redo app build caching

jmarrec:
I noticed an issue with how we cached the entire build directory...
* In most cases it doesn’t speed up anything
* Github limits caching to 10GB. We have 4 runners that cache about 3 to 4 GB, so we constantly exceed it
* We often pick up an outdated OS SDK installer

I’ve done some semi “clever” caching strategies now:

* Cache the conan cache. The key for the cache is computed from the autodetected conan profile (so specific to an OS, architecture, compiler version) + the hash of the ConanInstall.cmake
* Cache the build/Qt-install dir + OS SDK installer (but trash it later if it's not the exact one). This is dependent on the runner.os and the QT version we use
* Enable CCACHE on all platforms. This is dependent on the hash of the conan profile
    * I have a few caveats here: if the cache is hit, then it's not saved in post build. Fine for the conan cache, but probably want an update for the build stuff? ideally the cache would be updated constantly... I might revisit this. Note that cache is immutable, so that'd mean uploading a NEW cache entry

Build times improvements on a rebuild when the source hasn't changed:

* Ubuntu: 41 min -> 10 min
* macos-13 (where conan binaries needed to be built): 1h15 ->  27min  (shaving 23 min due to conan cache only)
* Windows: 1h15 -> 20 min (the compile step takes a penalty on initial ccache building, approx 1h versus 53 min without)
  

before:

image

after:

image

### DIFF
--- a/.github/workflows/app_build.yml
+++ b/.github/workflows/app_build.yml
@@ -157,6 +157,8 @@ jobs:
           echo CONAN_USER_HOME=${{ github.workspace }}/conan-cache >> $GITHUB_ENV
           echo CONAN_USER_HOME_SHORT=${{ github.workspace }}/conan-cache/short >> $GITHUB_ENV
           echo CONAN_PROFILE_DEFAULT=${{ github.workspace }}/conan-cache/.conan/profiles/default >> $GITHUB_ENV
+          CCACHE_DIR=${{ github.workspace }}/.ccache
+          echo CCACHE_DIR=$CCACHE_DIR >> $GITHUB_ENV
         fi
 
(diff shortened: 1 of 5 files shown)


jmarrec:
cd this workflow: without ccache hit (run 1) and with ccache hit (run 2): https://github.com/openstudiocoalition/OpenStudioApplication/actions/runs/7491867085/job/20399061453

jmarrec:
Ok all good, I even tested partial cache hit on https://github.com/openstudiocoalition/OpenStudioApplication/actions/runs/7494140697/job/20401388058 (the `restore-key` is indeed used when the `key` didn't match for the CCache and the Conan caches)

PR #682, opened 2024-02-26
Speed up controllers

### LINKED ISSUE #681: Dragging construction set from library is very slow
Issue overview
--------------
Dragging a construction set from the library is very slow and/or will crash the application.  Reported at https://unmethours.com/question/98295/construction-sets-not-loading-from-library/

## Expected Behavior
Dragging a construction set from the library should be fast

## Steps to Reproduce
1. Load an ASHRAE library file
2. Go to Constructions tab
3. Drag a PriSch construction set into the model

### Environment

 - Platform (Operating system, version): All

### Context
Cannot create a model

### Workaround
Potential workaround is to use 1.6.0 which does not have the library search feature

macumber:
Fixes #681 by moving optimizations in certain controllers to all controllers

### DIFF
--- a/src/openstudio_lib/ConstructionObjectVectorController.cpp
+++ b/src/openstudio_lib/ConstructionObjectVectorController.cpp
@@ -43,51 +43,27 @@
 #include <openstudio/utilities/idd/IddEnums.hxx>
 
 #include <QMessageBox>
-#include <QTimer>
-#include <QMutex>
 
 namespace openstudio {
 
 // ConstructionObjectVectorController
 
 ConstructionObjectVectorController::ConstructionObjectVectorController(QWidget* parentWidget)
-  : ModelObjectVectorController(), m_reportScheduled(false), m_reportItemsMutex(new QMutex()), m_parentWidget(parentWidget) {}
+  : ModelObjectVectorController(), m_parentWidget(parentWidget) {}
 
-ConstructionObjectVectorController::~ConstructionObjectVectorController() {
-  delete m_reportItemsMutex;
-}
-
-void ConstructionObjectVectorController::reportItemsLater() {
-  m_reportScheduled = true;
-
-  QTimer::singleShot(0, this, &ConstructionObjectVectorController::reportItems);
-}
-
-void ConstructionObjectVectorController::reportItems() {
-  if (!m_reportItemsMutex->tryLock()) {
-    return;
-  }
-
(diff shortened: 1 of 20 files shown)


src/openstudio_lib/OSItemList.hpp
```diff
@@ -105,7 +103,6 @@ class OSItemList : public OSItemSelector
   bool m_itemsDraggable;
   bool m_itemsRemoveable;
   OSItemType m_type;
-  bool m_dirty;
```
macumber, line 108: `m_dirty` wasn't actually controlling anything here, all that `refresh` was doing was to clear `m_dirty`

src/openstudio_lib/ConstructionObjectVectorController.cpp
```diff
@@ -43,51 +43,27 @@
...
 
 ConstructionObjectVectorController::ConstructionObjectVectorController(QWidget* parentWidget)
-  : ModelObjectVectorController(), m_reportScheduled(false), m_reportItemsMutex(new QMutex()), m_parentWidget(parentWidget) {}
+  : ModelObjectVectorController(), m_parentWidget(parentWidget) {}
 
-ConstructionObjectVectorController::~ConstructionObjectVectorController() {
-  delete m_reportItemsMutex;
-}
-
-void ConstructionObjectVectorController::reportItemsLater() {
```
macumber, line 60: Only a few `OSVectorController` derived classes were specialized to have the `reportItemsLater` slot which schedules a call to `OSVectorController::reportItems`.  This PR pulls that functionality into the `OSVectorController`.  Now `reportItems` will schedule a call to a `reportItemsImpl` method that will run once so we don't report (and redraw) the same objects multipe times per execute loop.

src/openstudio_lib/ModelObjectListView.hpp
```diff
@@ -60,15 +62,20 @@ class ModelObjectListController : public OSVectorController
  private:
   openstudio::IddObjectType m_iddObjectType;
   model::Model m_model;
-  bool m_showLocalBCL;
+  bool m_isLibrary;
```
macumber, line 65: `m_showLocalBCL` meant to show LocalBCL objects, basically just meant that this was a library controller, I thought `m_isLibrary` was more clear.

jmarrec (approved):
Not sure I follow everything but I trust your judgment.

PR #692, opened 2024-03-17
Upgrade to Qt 6.6.2 to fix #688

### LINKED ISSUE #688: OpenStudio Application unusable on MacOS14 with M2
Issue overview
--------------
Opening an existing model will cause the application to crash. Navigating the application will also cause it to crash.

## Current Behavior
When the application is first opened, and I navigate to file>open and select an existing model, the application will ask if I want to save the existing model. I select "Don't Save" (because this is a blank model after just opening the app) and the Application crashes. This happens with any model, even the built in example model from File>Examples>Example Model.

I tried to see if I could create a model and save, and after creating a floorspace, I navigated to the first screen (Site) and the application crashed.

I attempted to enable DEBUG log and followed the instructions provided, but am not very familiar with command line interface and received an error that the directory could not be found. This is after opening the Terminal within the OpenStudioApp folder in Finder.

## Expected Behavior
The Application should open a model file.

## Steps to Reproduce

Example model: https://github.com/NREL/OpenStudio/blob/develop/resources/Examples/compact_osw/files/seb.osm

1. Open OpenStudio Application
2. File>Open>"Any OpenStudio model" OR File>Examples>Example Model

## Possible Solution
OpenStudio Application 1.6. worked fine without any crashes.

### Environment

 - Platform (Operating system, version): MacOS14 on an M2 Macbook

macumber:
Upgrade to Qt 6.6.2 to fix #688

### DIFF
--- a/.github/workflows/app_build.yml
+++ b/.github/workflows/app_build.yml
@@ -28,7 +28,7 @@ env:
   CPACK_SOURCE_TXZ: OFF
   CPACK_SOURCE_TZ: OFF
   CPACK_SOURCE_ZIP: OFF
-  QT_VERSION: 6.6.1
+  QT_VERSION: 6.6.2
   # CPACK_BINARY_DEB: OS-SPECIFIC
   # CPACK_BINARY_IFW: OS-SPECIFIC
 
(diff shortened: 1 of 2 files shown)


jmarrec:
Oof just noticed this. I guess time to upgrade Qt on my machines :/

Qt 6.6.3 is going to be released in 3 days FYI.

PR #697, opened 2024-03-26
Fix #672 - Add description, modelerDescription, UID and version UID to bcl results

### LINKED ISSUE #672: BCL Viewer Human Readable Descriptions
Enhancement Request
--------------

It would be really helpful if the Readme files or human readable description and/or human readable description of modeling approach could be displayed when searching the BCL.  As it is, the data displayed is not very descriptive or helpful.  

This is also a problem with the BCL browser online.  Many of the descriptions are cut off; you must download the measure to actually figure out what it is.

## Detailed Description

Screenshot 2024-01-22 091829

jmarrec:
- Fix #672 
- Add description, modelerDescription, UID and version UID to bcl results

image

### DIFF
--- a/src/shared_gui_components/Component.cpp
+++ b/src/shared_gui_components/Component.cpp
@@ -507,8 +507,9 @@ void Component::createCompleteLayout() {
     }
     {
       auto* content = new QLabel(m_description);
-      mainLayout->addWidget(content);
       content->setWordWrap(true);
+      content->setOpenExternalLinks(true);
+      mainLayout->addWidget(content);
     }
   }
 
@@ -520,8 +521,9 @@ void Component::createCompleteLayout() {
     }
     {
       auto* content = new QLabel(m_modelerDescription);
-      mainLayout->addWidget(content);
       content->setWordWrap(true);
+      content->setOpenExternalLinks(true);
+      mainLayout->addWidget(content);
     }
   }
 
--- a/src/shared_gui_components/Component.cpp
+++ b/src/shared_gui_components/Component.cpp
@@ -544,7 +544,7 @@ void Component::createCompleteLayout() {
     return tableWidget;
   };
 
-  auto addRowToTableWiget = [](QTableWidget* tableWidget, const QString& header, const QString& value) {
+  auto addRowToTableWidget = [](QTableWidget* tableWidget, const QString& header, const QString& value) {
     tableWidget->insertRow(tableWidget->rowCount());
 
     auto* item = new QTableWidgetItem(header);
@@ -601,7 +601,7 @@ void Component::createCompleteLayout() {
         display += temp.c_str();
       }
 
-      addRowToTableWiget(tableWidget, attribute.name().c_str(), display);
+      addRowToTableWidget(tableWidget, attribute.name().c_str(), display);
     }
 
     makeTableShowCompletely(tableWidget);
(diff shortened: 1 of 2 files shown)


src/shared_gui_components/Component.cpp
```diff
@@ -464,157 +481,170 @@ void Component::createAbridgedLayout() {
...
+    {
+      auto* label = new QLabel("Modeler Description");
+      label->setObjectName("H1");
+      mainLayout->addWidget(label);
+    }
+    {
+      auto* content = new QLabel(m_modelerDescription);
+      mainLayout->addWidget(content);
+      content->setWordWrap(true);
+    }
```
jmarrec, line 525: Alternative (`#include <QTextEdit>` needed atop)

```suggestion
    {
      auto* content = new QTextEdit(m_modelerDescription);
      content->setStyleSheet("background: #E6E6E6;");
      content->setFixedHeight(70);
      content->setAcceptRichText(false);
      content->setVerticalScrollBarPolicy(Qt::ScrollBarAsNeeded);
      content->setTabChangesFocus(true);
      content->setReadOnly(true);

      mainLayout->addWidget(content);
    }
```

image

image

src/shared_gui_components/Component.cpp
```diff
@@ -464,157 +479,172 @@ void Component::createAbridgedLayout() {
...
-  tableWidget->setHorizontalScrollBarPolicy(Qt::ScrollBarAlwaysOff);
-  tableWidget->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Fixed);
-  tableWidget->horizontalHeader()->setSectionResizeMode(QHeaderView::ResizeToContents);
-  tableWidget->horizontalHeader()->setStretchLastSection(true);
-
-  mainLayout->addWidget(tableWidget);
+  // Description
+  if (!m_description.isEmpty()) {
+    {
+      auto* label = new QLabel("Description");
```
macumber, line 502: Might want to show the section even if it's empty?

src/shared_gui_components/Component.cpp
```diff
@@ -464,157 +479,172 @@ void Component::createAbridgedLayout() {
...
+        display.setNum(attribute.valueAsDouble());
+      } else if (type == AttributeValueType::Integer) {
+        display.setNum(attribute.valueAsInteger());
+      } else if (type == AttributeValueType::Unsigned) {
+        display.setNum(attribute.valueAsUnsigned());
+      } else if (type == AttributeValueType::String) {
+        display = attribute.valueAsString().c_str();
+      } else if (type == AttributeValueType::AttributeVector) {
+        AttributeVector attributeVector = attribute.valueAsAttributeVector();
+        // TODO handle this case
```
macumber, line 594: Looks like `attribute.toString` might work for the vector case, might be kind of ugly but I think this is rare.

src/shared_gui_components/Component.cpp
macumber: I think we'd need scroll bars here, either for each section like this or one scroll bar for the whole preview.

macumber:
Nice work!

src/shared_gui_components/Component.cpp
jmarrec: I think it looks kinda weird when I do that

src/shared_gui_components/Component.cpp
jmarrec: It wasn't there before and no one complained, so I didn't change it :)

toJSONString would be an optional, or just managed it by looping.

src/shared_gui_components/Component.cpp
macumber: Sounds good

src/shared_gui_components/Component.cpp
macumber: Yeah, I kind of doubt anyone is using AttributeVector

PR #699, opened 2024-03-26
Install openstudio to bin directory so it can be found relative to lib

### LINKED ISSUE #695: Error when running Create Typical Building from Model Measures
Issue overview
--------------
Error when running create typical building measures, specifically Create Typical Building from Model and Create DOE Prototype Building.

## Current Behavior
The above measures result in an error when running withing OpenStudio Application, however, no error result if running the workflow file created from the Application directly in CLI.

## Expected Behavior
Create Typical Building Model measure should create a building with loads from the Space Type and Construction Set Wizard.

## Steps to Reproduce

[ModelTest.zip](https://github.com/openstudiocoalition/OpenStudioApplication/files/14742526/ModelTest.zip)

1. Create a box model
2. Run measures in succession: 1. Space Type and Construction Set Wizard, 2. Create Typical Building from Model.

## Possible Solution

Running the OpenStudio Application generated workflow file in the CLI results in NO error.

## Details
Full error text file attached, but the specific portion where the error seems to occur is here. Running this same command but inserting "openstudio" in the "" runs fine with no error:

DEBUG - Running with OS 2.x WorkflowJSON.
[openstudio.standards.command] Error running command: '"" run -w "/private/var/folders/sc/plzbznf500sbbdjc5hppgxl40000gn/T/osmodel-9898-b823-4fdf-acbe-1711363291-0/resources/run/001_CreateTypicalBuildingFromModel/SRvt/in.osw"'
[openstudio.standards.command] stdout:
[openstudio.standards.command] stderr: sh: : command not found
[openstudio.standards.command] command's modified ENV:

### Environment

 - Platform (Operating system, version): MacOS14

### Context
Error File:  [ModelTest_Debug.txt](https://github.com/openstudiocoalition/OpenStudioApplication/files/14742479/ModelTest_Debug.txt)

macumber:
Install openstudio to bin directory so it can be found relative to lib, fixes #695

Basically, the openstudio CLI loads the libs from the lib folder, then getOpenStudioCLI looks for the CLI based on the libs paths.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -749,10 +749,10 @@ install(DIRECTORY "${openstudio_ROOT_DIR}/Python" DESTINATION "." COMPONENT "Pyt
 # TODO: at 3.5.0, the lib/ directory on UNIX (only UNIX) contains the libopenstudiolib.so, but also the librubyengine.so, libpythonengine.so
 # install(DIRECTORY "${openstudio_ROOT_DIR}/lib" DESTINATION "." COMPONENT "OpenStudioApp" USE_SOURCE_PERMISSIONS)
 
-if(NOT APPLE)
-  # On Mac, in openstudio_app/CMakeLists.txt we've already copied the CLI to the OpenStudio.app/Contents/MacOS
-  install(IMPORTED_RUNTIME_ARTIFACTS openstudio::openstudio DESTINATION bin COMPONENT "CLI")
-endif()
+# On Mac, in openstudio_app/CMakeLists.txt we've already copied the CLI to the OpenStudio.app/Contents/MacOS
+# However, it is still needed in the bin directory so it can be found relative to the files in lib, this occurs
+# when a measure tries to use getOpenStudioCLI, see #695
+install(IMPORTED_RUNTIME_ARTIFACTS openstudio::openstudio DESTINATION bin COMPONENT "CLI")
 install(IMPORTED_RUNTIME_ARTIFACTS openstudio::openstudiolib DESTINATION ${LIB_DESTINATION_DIR} COMPONENT "CLI")
 install(IMPORTED_RUNTIME_ARTIFACTS openstudio::rubyengine DESTINATION ${LIB_DESTINATION_DIR} COMPONENT "CLI")
 install(IMPORTED_RUNTIME_ARTIFACTS openstudio::pythonengine DESTINATION ${LIB_DESTINATION_DIR} COMPONENT "CLI")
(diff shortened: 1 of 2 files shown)


macumber:
@jmarrec this seems like the best thing to do in the short term.  We can try to find a better fix later but the layout in the OpenStudio Application install has to match what getOpenStudioCLI is expecting.  My thought was we should add an environment variable that the app could use to pass the path into getOpenStudioCLI

macumber:
@jmarrec I'll go ahead and merge this tomorrow unless you object

jmarrec:
I don't want to make a copy of the executable. It's big. We should symlink instead.

macumber:
I think the right way to fix this would be to move the content of the lib folder into the Mac bundle, the idea is you can move the bundle around and it still works.  However, the Mac bundle doesn't use a `bin`, `lib` layout.  The important thing is that the openstudio CLI be located relative to the lib files as expected here https://github.com/NREL/OpenStudio/blob/develop/src/utilities/core/ApplicationPathHelpers.cxx.in

jmarrec:
Tested locally on my mac after installing the dmg:

```
$ pwd
/Applications/OpenStudioApplication-1.7.0-test/bin
$ ls -la
total 304
drwxr-xr-x   4 julien  admin     128 Mar 27 14:36 .
drwxr-xr-x  20 julien  admin     640 Mar 27 15:20 ..
-rwxr-xr-x   1 julien  admin  152432 Sep 27  2023 install_utility
lrwxr-xr-x   1 julien  admin      46 Mar 27 14:36 openstudio -> ../OpenStudioApp.app/Contents/MacOS/openstudio
$ ./openstudio --version
3.7.0+d5269793f1
```

Running the test_os_cli from #695 

Class Name: TestGetOpenStudioCLI
Measure Type: ModelMeasure
Measure Language: Ruby
Current step has 0 arguments
argumentMap={}
Searching for CLI in '/Applications/OpenStudioApplication-1.7.0-test/lib/../openstudio'
Searching for CLI in '/Applications/OpenStudioApplication-1.7.0-test/lib/../bin/openstudio'
Found CLI at '/Applications/OpenStudioApplication-1.7.0-test/OpenStudioApp.app/Contents/MacOS/openstudio'
OpenStudio.getOpenStudioCLI=/Applications/OpenStudioApplication-1.7.0-test/OpenStudioApp.app/Contents/MacOS/openstudio
Step Result: Success
Finished applying OpenStudio Measures.

macumber:
@jmarrec this works for me, changes to fully self-contain the bundle would be too high risk right now

PR #700, opened 2024-03-27
Correctly remove old measures when downloading new ones

### LINKED ISSUE #696: Local copy of BCL folder gets wiped/overwritten
Issue overview
--------------

Often, when I am adding a measure from the BCL online, it asks me to update my local files with the latest.  When I click ok, it completely wipes all of my local BCL files from the BCL folder and I am left with just the one or two that I just downloaded.  I then have to search for and redownload all of the measures that I previously had.

## Steps to Reproduce

1.
2.
3.
4.

### Environment

 - Platform (Operating system, version): Windows 11
image

### Context

This is a major hassle.

macumber:
Properly remove components and measures of the old version, add missing pngs to qrc, don't remove all old measures after updating measures

Fixes #696 

Behavior will be further improved with https://github.com/NREL/OpenStudio/pull/5127

### DIFF
--- a/src/openstudio_lib/OSDocument.cpp
+++ b/src/openstudio_lib/OSDocument.cpp
@@ -1420,6 +1420,11 @@ size_t OSDocument::removeOutdatedLocalComponents(const std::string& uid, const s
   // }
   // return result;
 
+  auto components = getLocalComponents();
+  if (components.empty()) {
+    return {};
+  }
+
   // Not empty, we do have a localbcl
   components.erase(std::remove_if(components.begin(), components.end(),
                                   [&uid, &currentVersionId](const auto& component) {
@@ -1434,6 +1439,17 @@ size_t OSDocument::removeOutdatedLocalComponents(const std::string& uid, const s
 
 size_t OSDocument::removeOutdatedLocalMeasures(const std::string& uid, const std::string& currentVersionId) const {
   // TODO: when https://github.com/NREL/OpenStudio/pull/5129 is merged, we can just call it
+  // size_t result = 0;
+  // if (m_haveLocalBCL) {
+  //   try {
+  //     result = LocalBCL::instance().removeOutdatedLocalMeasures(uid, currentVersionId);
+  //   } catch (const std::exception& e) {
+  //     LOG(Error, "Cannot access local BCL: " << e.what());
+  //     m_haveLocalBCL = false;
+  //   }
+  // }
+  // return result;
+
   auto measures = getLocalMeasures();
   if (measures.empty()) {
     return {};
--- a/src/openstudio_lib/OSDocument.cpp
+++ b/src/openstudio_lib/OSDocument.cpp
@@ -1380,6 +1380,25 @@ boost::optional<BCLComponent> OSDocument::getLocalComponent(const std::string& u
   return result;
 }
 
+size_t OSDocument::removeOutdatedLocalComponents(const std::string& uid, const std::string& currentVersionId) {
+  size_t result = 0;
+  if (m_haveLocalBCL) {
+    try {
+      for (auto& oldComponent : LocalBCL::instance().components()) {
+        if ((oldComponent.uid() == uid) && (oldComponent.versionId() != currentVersionId)) {
+          if (LocalBCL::instance().removeComponent(oldComponent)) {
+            ++result;
+          }
+        }
+      }
+    } catch (const std::exception& e) {
+      LOG(Error, "Cannot access local BCL: " << e.what());
+      m_haveLocalBCL = false;
+    }
+  }
+  return result;
+}
+
 boost::optional<BCLMeasure> OSDocument::getLocalMeasure(const std::string& uid) const {
   boost::optional<BCLMeasure> result;
   if (m_haveLocalBCL) {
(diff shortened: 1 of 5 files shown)


macumber:
@jmarrec I'll test this some more before merging, let me know if you have any objections, it is in code near some of your changes.

src/openstudio_lib/OSDocument.cpp
```diff
@@ -1406,6 +1406,19 @@ boost::optional<BCLMeasure> OSDocument::getLocalMeasure(const std::string& uid,
   return result;
 }
 
+std::vector<BCLComponent> OSDocument::getLocalComponents() const {
+  std::vector<BCLComponent> result;
+  if (m_haveLocalBCL) {
```
jmarrec, line 1411: (I'm always thrown off by this ("Why would local bcl not be available") until I remember why we added it... )

src/openstudio_lib/OSDocument.cpp
```diff
@@ -1406,6 +1406,19 @@ boost::optional<BCLMeasure> OSDocument::getLocalMeasure(const std::string& uid,
   return result;
 }
 
+std::vector<BCLComponent> OSDocument::getLocalComponents() const {
```
jmarrec, line 1409: I don't see why we need this.

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -287,10 +287,12 @@ void BuildingComponentDialogCentralWidget::comboBoxIndexChanged(const QString& t
...
-    // remove old component
-    boost::optional<BCLComponent> oldComponent = OSAppBase::instance()->currentDocument()->getLocalComponent(component->uid());
-    if (oldComponent && oldComponent->versionId() != component->versionId()) {
-      LocalBCL::instance().removeComponent(*oldComponent);
+
+    // remove old components
+    for (auto& oldComponent : OSAppBase::instance()->currentDocument()->getLocalComponents()) {
+      if ((oldComponent.uid() == component->uid()) && (oldComponent.versionId() != component->versionId())) {
+        LocalBCL::instance().removeComponent(oldComponent);
+      }
```
jmarrec, line 295: Why is looping on all components as opposed to using getLocalComponent(uid) better/needed?!

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -306,11 +308,13 @@ void BuildingComponentDialogCentralWidget::measureDownloadComplete(const std::st
...
 
-    // remove old measure
-    boost::optional<BCLMeasure> oldMeasure = OSAppBase::instance()->currentDocument()->getLocalMeasure(measure->uid());
-    if (oldMeasure && oldMeasure->versionId() != measure->versionId()) {
-      LocalBCL::instance().removeMeasure(*oldMeasure);
+    // remove old measures
+    for (auto& oldMeasure : OSAppBase::instance()->currentDocument()->getLocalMeasures()) {
+      if ((oldMeasure.uid() == measure->uid()) && (oldMeasure.versionId() != measure->versionId())) {
+        LocalBCL::instance().removeMeasure(oldMeasure);
+      }
```
jmarrec, line 315: Same question

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
jmarrec: Oh, I think I see. getComponent(uid) returns the first it finds, is that it?

src/shared_gui_components/MeasureManager.cpp
```diff
@@ -981,9 +981,6 @@ void MeasureManager::checkForRemoteBCLUpdates() {
     int result = msg.exec();
     if (result == QMessageBox::Yes) {
       remoteBCL.updateMeasures();
-      for (auto& oldMeasure : oldMeasures) {
-        LocalBCL::instance().removeMeasure(oldMeasure);
-      }
```
jmarrec, line 986: So that's the actual fix? There's similar code a dozen lines below though.

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
jmarrec: I want to move this logic inside OSDocument where we do check if local bcl is available.

macumber: Yeah, we should just add new methods to OSDocument similar to what you added in https://github.com/NREL/OpenStudio/pull/5129.  We can update those to call the new methods when they become available.

src/shared_gui_components/MeasureManager.cpp
macumber: If `remoteBCL.updateMeasures();` works correctly then it will remove the outdated measures.  However, that wasn't working reliably due to the issue you fixed in https://github.com/NREL/OpenStudio/pull/5129.  My guess is that I was seeing a case where the old measures weren't removed so I added this code in.  However, this code fails when the new measures aren't fetched for some reason, the old ones will just be cleared.  I assumed that is what was happening in the bug report.

PR #713, opened 2024-05-08
Implement support for further grouping HVAC library by category 

### LINKED ISSUE #123: Organize HVAC library by category
https://openstudio.uservoice.com/forums/308629-openstudio/suggestions/10473030-organize-hvac-library-by-category

Since 1.9.0, the quantity of available library objects when viewing the "HVAC System" makes it really hard to find an object. It'd be great to categorize the objects, for example following the EnergyPlus "Groups" (in the I/O documentation, or when using IDF editors)

You'd have a group for "Setpoint Managers", with all possible setpoints managers under it. One for "Fans" which all fans under it, one for Pumps, one for Condenser Equipment (Cooling Towers, Fluid Coolers, etc)

manuvarkey:
1. Implement `OSCategoryPlaceholder` class for further grouping inside `OSCollapsibleItemList`
2. Modification to search bar to use category in addition to type and name.
3. Resolves #123


src/openstudio_lib/OSCollapsibleItemList.cpp
```diff
@@ -167,6 +168,12 @@ void OSCollapsibleItemList::addCollapsibleItem(OSCollapsibleItem* collapsibleIte
   connect(collapsibleItem, &OSCollapsibleItem::openLibDlgClicked, this, &OSCollapsibleItemList::openLibDlgClicked);
 }
 
+void OSCollapsibleItemList::addCategoryPlaceholderItem(OSCategoryPlaceholder* categoryPlaceholderItem) {
+
+  m_placeholderItems.push_back(categoryPlaceholderItem);
+  m_vLayout->insertWidget(0, categoryPlaceholderItem);
+}
```
jmarrec, line 175: @macumber This makes me realize we have a `std::vector<OSCollapsibleItem*> m_collapsibleItems;` but our `void OSCollapsibleItemList::addCollapsibleItem(OSCollapsibleItem* collapsibleItem)` does not add to it... even though other functions loop on it...

jmarrec (commented):
Alright, this is **neat**, thanks a lot @manuvarkey 

I do question whether we shouldn't use an actual OSCollapsibleItemHeader directly, so we can expand it or not.

Here is a demo of the current code:

Peek 2024-05-13 15-25

src/openstudio_lib/OSCollapsibleItemList.cpp
macumber: Hmm, yeah looks like we never add anything to `m_collapsibleItems` so it's kind of useless.

macumber:
We did discuss adding a second level of expanding headers in https://github.com/openstudiocoalition/OpenStudioApplication/issues/123 but I steered @manuvarkey towards the solution he implemented because I thought the multiple levels of collapsable headers would get complicated.  

I updated the code to compile, changed the style of the placeholder, and added categories to all the tabs.  What do you think?

image

jmarrec:
The style looks great. And I can understand the rationale behind too much grouping = confusing.

LGTM, feel free to merge at will!

src/openstudio_lib/OSCategoryPlaceholder.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+namespace openstudio {
+
+OSCategoryPlaceholder::OSCategoryPlaceholder(const std::string& text, QWidget* parent) : QWidget(parent) {
+  setFixedHeight(40);
+  setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Fixed);
+  setObjectName("OSCategoryPlaceholder");
+
+  this->setProperty("style", "0");
+  this->setStyleSheet("QWidget#OSCategoryPlaceholder[style=\"0\"] { background-color: #95B3DE; border-bottom: 1px solid black; }");
```
jmarrec, line 49: More of a side note than anything, but I wonder when we should just inline styles versus using the qss stylesheet. @macumber do you have rules for this?

https://github.com/openstudiocoalition/OpenStudioApplication/blob/a9b492b6df93d9295e21ce4da46f1ad67788bf87/src/openstudio_lib/openstudiolib.qss#L989-L995

macumber: I'm always torn between trying to write documentation and rules for what exists vs burn it all down and start over.  I think it would make the most sense to have all styes in qss files, but that would require a sensible naming hierarchy everywhere.  It gets hard when a name changes and the styles don't apply to some tab anymore.

Issue #612, opened 2023-07-09
Show displayname is grid view


macumber:
Enhancement Request
--------------
For applications where displayname is different than the object name, it would be useful to see it in the OS app.  Not sure if it should be a separate column or something else.

https://unmethours.com/question/93379/display-names-of-items-imported-from-gbxml/

christophertindall:
This needs to be elevated above an "Enhancement Request".  The fix for [OpenStudio Issue 4457](https://github.com/NREL/OpenStudio/issues/4457) has broken our ability to use OpenStudio Application to modify OSM files generated by Revit Systems Analysis.  OpenStudio Application now only displays the gbXMLId values, which is not useful for navigating the model.  While we are waiting for this fix, can someone write an OpenStudio measure to replace the space gbXMLId values with the space displayName values?  Revit already verifies the displayName value is unique so it should be a simply measure to write.

jmarrec:
I understand your concern if this is breaking your workflow. That being said, I'd like to clarify something:

* Revit produces gbxml in a certain way. 
* OpenStudio SDK can handle the conversion between to/from gbxml/OSM. 
* OpenStudioApplication uses openstudio SDK and works on an OSM. 

From the point of view of the OpenStudioApplication: a model object has a name, we display it, that's about it. It works as expected and designed. We control neither how Revit produces gbxml nor the OpenStudio SDK's handling of gbxml.

So yes, this _is_ an enhancement request. And it is similar to 

* https://github.com/openstudiocoalition/OpenStudioApplication/issues/417

jmarrec:
the OSM has the gbxml's id as the name.

Here is some python code to replace it.

```python
import openstudio

model = openstudio.model.Model.load('model.osm').get()

for obj in model.modelObjects():
    if not obj.hasAdditionalProperties():
        continue
    displayName_ = obj.displayName()
    if not displayName_.is_initialized():
        continue
    name = obj.nameString()
    displayName = displayName_.get()
    gbXMLId_ = obj.gbXMLId()
    print(f"For {obj.briefDescription()}, setting name to '{displayName}'")
    if gbXMLId_.is_initialized():
        gbXMLId = gbXMLId_.get()
        if gbXMLId != name:
            print(f"and replacing gbXMLId '{gbXMLId}' with '{name}'")
    obj.setName(displayName)
    obj.setGBXMLId(name)

model.save('replaced.osm', True)
```

same in ruby

```ruby
require 'openstudio'

[...]

PR #715, opened 2024-05-23
#612 - Display Name

jmarrec:
Just a proof of concept for now. Just on Space tab.

* #612

image

I suppose we should have a toggle switch somewhere in the preferences...


macumber:
That looks great, I like a toggle but maybe on the tab itself?

jmarrec:
Theoretically speaking, this applies to space, surfaces, shading surfaces, thermal zones, etc etc.
A toggle on each of these subtabs would be a hassle no?

src/openstudio_lib/OSDocument.cpp
```diff
@@ -441,6 +444,7 @@ void OSDocument::createTab(int verticalId) {
   m_verticalId = verticalId;
 
   bool isIP = m_mainWindow->displayIP();
+  bool displayAdditionalProps = m_mainWindow->displayAdditionalProps();
```
macumber, line 447: Does `toggleDisplayAdditionalPropsClicked` get emitted somewhere on startup if it is set?

src/openstudio_lib/FacilityShadingGridView.cpp
```diff
@@ -498,6 +507,22 @@ void FacilityShadingGridController::addColumns(const QString& category, std::vec
                           boost::optional<std::function<void(model::ShadingSurface*)>>(
                             std::function<void(model::ShadingSurface*)>([](model::ShadingSurface* t_ss) { t_ss->remove(); })),
                           boost::optional<std::function<bool(model::ShadingSurface*)>>(), DataSource(allShadingSurfaces, true));
+      } else if (field == DISPLAYNAME) {
+        addLoadNameColumn(Heading(QString(DISPLAYNAME), false, false),                                        // heading
```
macumber, line 511: Maybe just `addValueEditColumn` would work here?  Or `addNameLineEditColumn`?  To show the value without the icon?

src/openstudio_lib/OSDocument.cpp
jmarrec: It's read from settings and passed around. I dont think it needs to be emitted

src/openstudio_lib/FacilityShadingGridView.cpp
jmarrec: Yeah probably. I'll look into it

PR #727, opened 2024-08-24
Codesign on macOS

jmarrec:
# Included works

The absolutely required bit is 

* Codesign an notarize the DMG installer itself (this is a self extractable executable)
    * This required patching QtIFW and creating binaries
        * https://github.com/jmarrec/QtIFW-fixup/releases/tag/v5.0.0-dev-with-fixup 
        * Trying to upstream them in https://codereview.qt-project.org/c/installer-framework/installer-framework/+/568786
       
The rules are going to be tightened in the fall when macOS Sequoia releases: https://9to5mac.com/2024/08/06/macos-sequoia-makes-it-harder-to-run-apps-that-dont-follow-apples-security-rules/

In addition, I chose to also codesign and harden the inner apps, libraries, and executables. This will help for the tar.gz one probably.

* Codesign OpenStudio.app itself
* Codesign all other libraries and executables (eg: Radiance, EnergyPlus, Python sos, etc)

I also added some QA/QC scripts that will, on GHA

* On the build machine:
     * it will check that the inner _CPack_Pacages binaries and executables are properly codesigned for IFW and TGZ
     * for the TGZ one, it checks whether all libs can be resolved (the script uses otool and some logic to check if everything is fine)
* On a test machine:
     * It downloads the dmg artifact,  ensures that is properly codesigned and notarized
     * Mount the dmg
     * check the installer is properly codesigned
     * performs a silent install
     * unmount dmg
     * Check the signature of all the inner binaries and executables that have been installed
     * otool to ensure all libraries are resolved (and output a JSON file for debugging / manual checking)
     * Performs an E+ simulation with a PythonPlugin to ensure that works fine

# Demonstration

## Before signing

You get a an eject warning, and you have to right click > Open if you want to install it.

macos_unsigned_installer

## With signing

macos_signed_installer


.github/workflows/app_build.yml
```diff
@@ -628,3 +698,102 @@ jobs:
         /bin/rm OpenStudioApplication-*${{ env.COMPRESSED_EXT }} || true
         /bin/rm OpenStudioApplication-*${{ env.BINARY_EXT }} || true
         ls OpenStudioApplication-* || true
+
+
+  test_package:
```
macumber, line 703: Is this guaranteed to run on the same machine as the build above for non-self-hosted runners?

macumber: Or it doesn't matter because you are re-downloading the artifact.

jmarrec: The idea was the opposite. I wanted to do the testing bit on a separate machine. Which is now what I've done including for arm64 by using a GHA hosted macos-14 runner (which is arm64).

FWIW, I suppose at some point the overhead of maintaining the self hosted mac mini will make us switch to a a GHA hosted one, even though it has less CPU power.

jmarrec:
There's an intermittent failure in the packaging step, that has been prexisting for a while, but which is annoying the crap out of me during all this testing.

There's a create_symlink failing sometimes, but restarting the run usually makes it work...
https://github.com/openstudiocoalition/OpenStudioApplication/actions/runs/10575968626/job/29300787775#step:18:2721

https://github.com/openstudiocoalition/OpenStudioApplication/blob/bcae7d2dfb684f8d81882c632d48e27945b62bd5/src/openstudio_app/CMakeLists.txt#L446-L484

jmarrec:
Ok, finally all green, aside from that nagging intermittent failure. Renabling all workflows now. Assuming I can pass CI just fine, this can drop.

.github/workflows/app_build.yml
macumber: Yup totally agree, last I checked the MAC arm runner was either not available or really expensive or something.  It will be great when we can deprecate the self hosted one. 

Testing on clean runners makes a lot of sense.

.github/workflows/app_build.yml
```diff
@@ -190,33 +199,65 @@ jobs:
...
         ccache --zero-stats || true
 
+    - name: "Configure for codesigning"
+      if: runner.os == 'macOS'
+      run: |
+        set -x
+        cd $RUNNER_TEMP
+        mkdir codesigning && cd codesigning
+        # ----- Create certificate files from secrets base64 -----
+        echo "${{ secrets.MACOS_DEVELOPER_ID_APPLICATION_CERTIFICATE_P12_BASE64 }}" | base64 --decode > certificate_application.p12
```
macumber, line 216: Is this certificate going to expire every year?  Maybe we can set up a reminder to renew it two weeks before it expires?

.github/workflows/app_build.yml
```diff
@@ -628,3 +701,101 @@ jobs:
         /bin/rm OpenStudioApplication-*${{ env.COMPRESSED_EXT }} || true
         /bin/rm OpenStudioApplication-*${{ env.BINARY_EXT }} || true
         ls OpenStudioApplication-* || true
+
+
+  test_package_macos:
```
macumber, line 706: This is cool, we should consider doing this for the other platforms too!

CMake/install_codesign_script_OpenStudioApp.cmake
```diff
@@ -0,0 +1,307 @@
...
+if(NOT CPACK_CODESIGNING_DEVELOPPER_ID_APPLICATION)
+  message(FATAL_ERROR "CPACK_CODESIGNING_DEVELOPPER_ID_APPLICATION is required")
+endif()
+
+if(NOT CPACK_CODESIGNING_MACOS_IDENTIFIER)
+  message(FATAL_ERROR "CPACK_CODESIGNING_MACOS_IDENTIFIER is required")
+endif()
+
+
+set(FILES_TO_SIGN
```
macumber, line 101: Is there anyway to get this list of executables from CMake?  Or do a glob after the build?

CMakeLists.txt
```diff
@@ -922,6 +931,38 @@ elseif(WIN32)
...
+    set(CLI_FILES_TO_SIGN
+      lib/libopenstudiolib.dylib
+      lib/libpythonengine.so
+      lib/librubyengine.so
+      bin/install_utility
+    )
+    install(CODE "set(FILES_TO_SIGN \"${CLI_FILES_TO_SIGN}\")" COMPONENT "CLI")
+    install(CODE "set(CPACK_CODESIGNING_DEVELOPPER_ID_APPLICATION \"${CPACK_CODESIGNING_DEVELOPPER_ID_APPLICATION}\")" COMPONENT CLI)
+    install(CODE "set(CPACK_CODESIGNING_MACOS_IDENTIFIER \"${CPACK_CODESIGNING_MACOS_IDENTIFIER}\")" COMPONENT CLI)
+    install(SCRIPT "${CMAKE_CURRENT_LIST_DIR}/CMake/install_codesign_script.cmake" COMPONENT CLI)
```
macumber, line 949: Does this signing happen at install time on the users computer?  Does this leak our certificate information into the installer?

macumber:
Wow @jmarrec this is a lot of work, very impressive!  I left some comments/questions

.github/workflows/app_build.yml
jmarrec: I'll check tomorrow but I think only in 2026 or 2027

CMakeLists.txt
jmarrec: It happens in the cpack staging area. So build/_Cpack_Packages/darwin/<tgz or ifw>

So no, no leaking

CMake/install_codesign_script_OpenStudioApp.cmake
jmarrec: Yes. But that is complicated enough for now and will work for the next 6months.

I'm not sure if I  can do pure cmake glob and find executables or libraries, but we could make a python script similar to the verify_signature I created and call that from cmake.
We'll need to have python as a build dependency but since we use conan already that's not a problem

jmarrec: Will work until the next time we upgrade os sdk to be clear. And then the signature testing I set on CI will tell us when it breaks and which files aren't signed properly

.github/workflows/app_build.yml
jmarrec: Different runners at least. The gha ones aren't vanilla at all, they have a lot of preinstalled dev tools.
The only clean one you can get is on linux because you can run on a vanilla docker image for ubuntu.
I've put in a suggestion to github to add vanilla runners but I don't expect they will

PR #731, opened 2024-08-29
Fix #162 - Display ScheduleConstant, ScheduleCompact and ScheduleFile

### LINKED ISSUE #162: Allow visualization of Schedule:Compact and Schedule:Constant on Schedules tab in OS App
Understanding that they might not be editable users are concerned if they cannot see their schedules in the model.  Kristin was concerned that all of her schedule compacts had been removed after importing model from idf.  

Alternatives are to show schedules in read only or convert all schedules to schedule ruleset

jmarrec:
# ScheduleConstant

Easy peasy

image

# ScheduleFile

Creating is possible, will ask for the external file on disk.

Added a lot of usability there, will display the content of the external file and show what data you're selecting, warn you if not matching, etc

image

# ScheduleCompact

Displaying the object as is. Not possible to create, will tell you to use a ScheduleRuleset instead

image

# Demo

### DIFF
--- a/src/openstudio_lib/ScheduleOthersController.cpp
+++ b/src/openstudio_lib/ScheduleOthersController.cpp
@@ -69,8 +69,8 @@ void ScheduleOthersController::onAddObject(const openstudio::IddObjectType& iddO
         if (e_) {
           model::ScheduleFile(*e_, 1, 1);
         }
-        break;
       }
+      break;
     }
     default:
       LOG_FREE_AND_THROW("ScheduleOthersController", "Unknown IddObjectType '" << iddObjectType.valueName() << "'");
--- a/src/openstudio_lib/ScheduleOthersController.cpp
+++ b/src/openstudio_lib/ScheduleOthersController.cpp
@@ -59,7 +59,6 @@ void ScheduleOthersController::onAddObject(const openstudio::IddObjectType& iddO
       message.setText("Creation of Schedule:Compact is not supported, you should use a ScheduleRuleset instead");
       message.exec();
       return;
-      break;
     }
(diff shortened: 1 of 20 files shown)


macumber (approved):
This is awesome, thanks @jmarrec!

src/openstudio_lib/ScheduleOthersController.cpp
```diff
@@ -0,0 +1,109 @@
...
+      QString selfilter = tr("CSV Files(*.csv)");
+      QString fileName = QFileDialog::getOpenFileName(this->subTabView(), tr("Select External File"), QDir::homePath(),
+                                                      tr("All files (*.*);;CSV Files(*.csv);;TSV Files(*.tsv)"), &selfilter);
+      if (!fileName.isNull()) {
+        boost::optional<model::ExternalFile> e_ = model::ExternalFile::getExternalFile(this->model(), fileName.toStdString());
+        if (e_) {
+          model::ScheduleFile(*e_, 1, 1);
+        }
+        break;
+      }
```
macumber, line 73: @jmarrec can you move the break down one level, right now if you cancel out of the file dialog the switch statement falls through into the default section and calls `LOG_FREE_AND_THROW`

jmarrec: good catch thanks

src/openstudio_lib/ScheduleOthersController.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
```
jmarrec, line 74:
```suggestion
      }
      break;
    }
```

PR #730, opened 2024-08-28
Fix #708 - Improve SetpointManager views: support more types, add plots where applicable, and support PlantLoop

jmarrec:
* Fix #708

PR #733, opened 2024-09-03
Fix #671 - Mark Python Measures as in Error if use classic CLI to run the simulation

### LINKED ISSUE #671: Mark Python measures as in error if using classic CLI
Enhancement Request
--------------
Classic CLI cannot run Python measures.  If Python measures are in the workflow, add error badge if using the Classic CLI along with a message saying "Cannot use Python measures with the Classic CLI".

Should also hide Python measures in library if the Measure Manager is launched with the Classic CLI.

jmarrec:
* Fix #671 
 
This will require you to change tab and come back after an update to your preferences, but that's ok because this is a temporary thing anyways, as the OpenStudio SDK has deprecated the classic CLI and will remove it soonish.

### DIFF
--- a/src/shared_gui_components/LocalLibraryController.cpp
+++ b/src/shared_gui_components/LocalLibraryController.cpp
@@ -521,7 +521,11 @@ QWidget* LibraryItemDelegate::view(QSharedPointer<OSListItem> dataSource) {
     // Name
 
     widget->label->setText(libraryItem->displayName());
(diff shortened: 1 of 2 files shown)


jmarrec:
Workflow failed because the build cache from develop was retrieved and this does not have QtCharts. Delete the caches now, will trigger rebuild (develop failed too since last merge of #730)

PR #754, opened 2024-09-16
Add an Ubuntu-2404 runner

jmarrec:


macumber (approved):
Looks good, my only question is if we should drop 20.04?  I’ll leave that up to you

jmarrec:
I was on 20.04 until last week, so my guess is that it's premature. We'll be forced to do it at some point when GHA removes the runner, but until then (and assuming E+ and OpenStudio both continue to do so), I think it's better to leave it in.

PR #759, opened 2024-10-09
Fix #758 - Properly encode path to QUrl when trying to display results

### LINKED ISSUE #758: Results tab shows Your file couldn’t be accessed
Issue overview
--------------
Apologies if this issue is reported before. I was not able to find anything on the account. When the OpenStudio file name has a special character (e.g., `#`), the results tab does not show the results.

## Current Behavior
When the file name has the special character, the following error message is shown:

```
Your file couldn’t be accessedIt may have been moved, edited, or deleted.
ERR_FILE_NOT_FOUND
```

Screenshot 2024-10-04 at 5 16 40 PM

## Expected Behavior
The results table should show the EnergyPlus and OpenStudio results similar to the image below:
Screenshot 2024-10-04 at 5 20 03 PM

## Steps to Reproduce
1. Just add `#` to any OpenStudio file name.

## Possible Solution
Remove `#` from the file name.

### Environment
I have seen this error before in different platforms and older version of OpenStudio but never reported it. The test is conducted on:

 - MAC 15.0.1 (24A348)
 - Version of OpenStudioApplication: 1.7

### Context
Recommended students to remove the `#` or see the reports folder.

Here is a [link](https://www.dropbox.com/scl/fi/tqq98711cq93s9uwor4po/Test-with-hashtag.zip?rlkey=lnttbe9avjma44uk4morh0gak&dl=0) to a sample model.

jmarrec:
My two cents: NEVER USE special characters for files or directory names. Nor space. 

Limit all file names to numbers ([0-9]) letters ([a-zA-Z]), and "-", "_" and "." You'll live a a much happier life generally.

### DIFF
--- a/src/openstudio_lib/ResultsTabView.cpp
+++ b/src/openstudio_lib/ResultsTabView.cpp
@@ -301,7 +301,6 @@ void ResultsView::populateComboBox(const std::vector<openstudio::path>& reports)
     QString fullPathString = toQString(report);
 
     QFile file(fullPathString);
-    fullPathString.prepend("file:///");
 
     if (openstudio::toString(report.filename()) == "eplustbl.html" || openstudio::toString(report.filename()) == "eplustbl.htm") {
 
@@ -360,7 +359,7 @@ void ResultsView::comboBoxChanged(int index) {
 
   m_progressBar->setError(false);
 
-  QUrl url(filename);
+  QUrl url = QUrl::fromLocalFile(filename);
   m_view->load(url);
 }
 
--- a/src/model_editor/test/Utilities_GTest.cpp
+++ b/src/model_editor/test/Utilities_GTest.cpp
@@ -123,7 +123,7 @@ TEST_F(ModelEditorFixture, MorePath_Conversions) {
     QString expectedUrl;
   };
 
-  std::vector<PathTestCase> testCases = { 
+  std::vector<PathTestCase> testCases = {
     {"C:\\Users\\Test\\eplustbl.html", "C:/Users/Test/eplustbl.html", "file:///C:/Users/Test/eplustbl.html"},
     {"C:/Users/Test/eplustbl.html", "C:/Users/Test/eplustbl.html", "file:///C:/Users/Test/eplustbl.html"},
     {"C:\\Users/Test/eplustbl.html", "C:/Users/Test/eplustbl.html", "file:///C:/Users/Test/eplustbl.html"},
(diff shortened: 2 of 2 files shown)


src/openstudio_lib/ResultsTabView.cpp
```diff
@@ -301,7 +301,6 @@ void ResultsView::populateComboBox(const std::vector<openstudio::path>& reports)
     QString fullPathString = toQString(report);
 
     QFile file(fullPathString);
-    fullPathString.prepend("file:///");
```
jmarrec, line 304: No need for that

src/openstudio_lib/ResultsTabView.cpp
```diff
@@ -360,7 +359,7 @@ void ResultsView::comboBoxChanged(int index) {
 
   m_progressBar->setError(false);
 
-  QUrl url(filename);
+  QUrl url = QUrl::fromLocalFile(filename);
```
jmarrec, line 362: because we use this.

https://doc.qt.io/qt-6/qurl.html#fromLocalFile

[qDebug](https://doc.qt.io/qt-6/qtlogging.html#qDebug)() << [QUrl](https://doc.qt.io/qt-6/qurl.html#QUrl)::fromLocalFile("/home/user/file.txt"); // QUrl("file:///home/user/file.txt")

macumber: Does `QUrl::fromLocalFile` escape bad characters (like `#`) in the path?  I don't see that in the documentation for the method.  I'll see if I can add a unit test for that.

macumber (approved):
Looks good, thanks!

src/openstudio_lib/ResultsTabView.cpp
jmarrec: It seems like it does indeed, since I no longer get the "file#1.osm" isssue.

I checked the source code to make sure, basically it sets the Scheme to fileScheme, and eventually the q_urlRecode is called:

https://github.com/qt/qtbase/blob/175569e1ec7f4d2ff7ace442c23c0bcdb86063d0/src/corelib/io/qurlrecode.cpp#L603-L639

jmarrec: Chain of call: 

https://github.com/qt/qtbase/blob/175569e1ec7f4d2ff7ace442c23c0bcdb86063d0/src/corelib/io/qurl.cpp#L3345-L3346

https://github.com/qt/qtbase/blob/175569e1ec7f4d2ff7ace442c23c0bcdb86063d0/src/corelib/io/qurl.cpp#L1111-L1115

https://github.com/qt/qtbase/blob/175569e1ec7f4d2ff7ace442c23c0bcdb86063d0/src/corelib/io/qurl.cpp#L808

jmarrec: And the characters handled are these ones: https://github.com/qt/qtbase/blob/175569e1ec7f4d2ff7ace442c23c0bcdb86063d0/src/corelib/io/qurl.cpp#L727-L738

PR #761, opened 2024-10-14
Fix #262 #263 - Handle Path Arguments (QFileDialog)

### LINKED ISSUE #262: Add file system browse control for measure path arguments
##  Expected Behavior
Path arguments of measures should allow opening filesystem finders

## Current Behavior
Only string inputs are allowed

### LINKED ISSUE #263: Cannot enter path arguments in apply measure now
## Current Behavior
Path arguments allow for users to enter a string value, when the measure is run `getPathArgumentValue` will fail

image

jmarrec:
* Fix #262 
* Fix #263

The OS SDK / BCL-gem implementation of the path arguments is incomplete and confusing, and they don't handle the makePathArgument  arguments:

*  `isRead`. what does that even mean? Is this assuming that this is for **reading** meaning it must be 1) a file (not a directory) and 2) it must exist?)
* `extension`

path_arguments

### DIFF
--- a/src/shared_gui_components/EditController.cpp
+++ b/src/shared_gui_components/EditController.cpp
@@ -7,6 +7,7 @@
 #include "EditView.hpp"
 #include "OSViewSwitcher.hpp"
 #include "WorkflowController.hpp"
+#include "../model_editor/Utilities.hpp"
 
 #include <openstudio/utilities/bcl/BCLMeasure.hpp>
 #include <openstudio/utilities/core/Assert.hpp>
@@ -251,6 +252,22 @@ InputController::InputController(EditController* editController, measure::OSArgu
             static_cast<void (InputController::*)(const QString&)>(&InputController::setValue));
 
     inputView = stringInputView;
+  } else if (m_argument.type() == measure::OSArgumentType::Path) {
+    auto* pathInputView = new PathInputView(m_argument.extension(), m_argument.isRead());
+
+    pathInputView->setName(m_argument.displayName(), m_argument.units(), m_argument.description());
+
+    if (m_argument.hasValue()) {
+      pathInputView->lineEdit->setText(QString::fromStdString(m_argument.valueAsString()));
+    } else if (m_argument.hasDefaultValue()) {
+      pathInputView->lineEdit->setText(QString::fromStdString(m_argument.defaultValueAsString()));
+    }
+
+    connect(pathInputView, &PathInputView::selectedPathChanged, this,
+            static_cast<void (InputController::*)(const openstudio::path&)>(&InputController::setValue));
+    connect(pathInputView->lineEdit, &QLineEdit::textEdited, [this](const QString& v) { this->setValue(toPath(v)); });
+
+    inputView = pathInputView;
   } else {
     inputView = new InputView();
   }
@@ -294,6 +311,20 @@ void InputController::setValueForIndex(int index) {
       m_argument.clearValue();
     } else {
       m_argument.setValue(value.toStdString());
+    }
+
+    m_editController->measureStepItem()->setArgument(m_argument);
+
+    inputView->setIncomplete(isArgumentIncomplete());
+  }
+}
+
+void InputController::setValue(const openstudio::path& p) {
+  if (isItOKToClearResults()) {
+    if (p.empty()) {
+      m_argument.clearValue();
+    } else {
+      m_argument.setValue(p);
     }
 
     m_editController->measureStepItem()->setArgument(m_argument);
--- a/src/shared_gui_components/EditView.cpp
+++ b/src/shared_gui_components/EditView.cpp
@@ -4,6 +4,8 @@
 ***********************************************************************************************************************/
 
 #include "EditView.hpp"
+#include "../model_editor/Utilities.hpp"
+
 #include <QVBoxLayout>
 #include <QHBoxLayout>
 #include <QLabel>
@@ -18,6 +20,8 @@
 #include <QCheckBox>
 #include <QWheelEvent>
 #include <QApplication>
+#include <QPushButton>
+#include <QFileDialog>
 
 #include <openstudio/utilities/core/Assert.hpp>
 
@@ -369,8 +373,6 @@ InputCheckBox::InputCheckBox() : QAbstractButton() {
   setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Fixed);
 }
 
-InputCheckBox::~InputCheckBox() = default;
-
 void InputCheckBox::setText(const QString& text) {
   m_label->setText(text);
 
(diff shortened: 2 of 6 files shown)


jmarrec:
here is my test measure

[example_path_argument.zip](https://github.com/user-attachments/files/17363128/example_path_argument.zip)


```
  # define the arguments that the user will input
  def arguments(model)
    args = OpenStudio::Measure::OSArgumentVector.new

    isRead = false
    extension = "CSV (*.csv);;Excel (*.xls *.xlsx);;All Files (*)"
    required = true
    modelDependent = false
    output_path = OpenStudio::Measure::OSArgument.makePathArgument('output_path', isRead, extension, required, modelDependent)
    output_path.setDisplayName('Output Path on Disk')
    args << output_path

    return args
  end

  # define what happens when the measure is run
  def run(model, runner, user_arguments)
    super(model, runner, user_arguments)  # Do **NOT** remove this line

    # use the built-in error checking
    if !runner.validateUserArguments(arguments(model), user_arguments)
      return false
    end

    # assign the user inputs to variables
    output_path = runner.getPathArgumentValue('output_path', user_arguments)

    puts "output_path=#{output_path}"

    File.write(output_path.to_s, "hello")

    # report final condition of model
    runner.registerFinalCondition("The output path is #{output_path}")

    return true
  end
```

src/shared_gui_components/MeasureManager.cpp
```diff
@@ -607,8 +607,10 @@ boost::optional<measure::OSArgument> MeasureManager::getArgument(const measure::
 
   } else if (type.value() == measure::OSArgumentType::Path) {
 
+    // TODO: aside from the fact that these arguments are weird / incorrectly named, neither the BCL-gem schema nor OS SDK actually handle them so
+    // they are not even inside the measure.xml
     bool isRead = argument.get("is_read", Json::Value(false)).asBool();
-    std::string extension = argument.get("extension", Json::Value("*")).asString();
+    std::string extension = argument.get("extension", Json::Value("All files (*)")).asString();
```
jmarrec, line 613: Default to something sensible for now

filed at https://github.com/NREL/OpenStudio/issues/5273

src/shared_gui_components/WorkflowController.cpp
```diff
@@ -501,6 +501,8 @@ void MeasureStepItem::setArgument(const measure::OSArgument& argument) {
       m_step.setArgument(argument.name(), argument.valueAsDouble());
     } else if (argument.type() == measure::OSArgumentType::Integer) {
       m_step.setArgument(argument.name(), argument.valueAsInteger());
+    } else if (argument.type() == measure::OSArgumentType::Path) {
+      m_step.setArgument(argument.name(), openstudio::toString(argument.valueAsPath()));
```
jmarrec, line 505: Without this you, you get valueAsString, and it gets quoted in the workflowJSON as 

```
    "my_path_argument": "\"/the/path/\""
```

and when you run, Ruby for eg can't understand it.

src/shared_gui_components/EditView.cpp
```diff
@@ -400,4 +402,80 @@ void InputCheckBox::setIncomplete(bool incomplete) {
...
+  selectPathButton->setFixedSize(30, 20);
+  selectPathButton->setObjectName("StandardGrayButton");
+
+  connect(selectPathButton, &QPushButton::clicked, this, &PathInputView::onSelectPathButtonClicked);
+  hLayout->addWidget(selectPathButton);
+
+  vLayout->addLayout(hLayout);
+
+  // TODO: sanitize the input?
+  m_extension = toQString(extension);
```
jmarrec, line 430: TODO here. Basically I think OS SDK should store it as a readable format, and we can reconstruct a proper Qt files Filter from it.

src/shared_gui_components/EditView.hpp
```diff
@@ -196,6 +198,35 @@ class InputCheckBox : public QAbstractButton
...
+ signals:
+  void selectedPathChanged(const openstudio::path& p);
+
+ private slots:
+  void onSelectPathButtonClicked();
+
+ private:
+  QLabel* nameLabel;
+  QString m_extension;
+  bool m_isRead;
```
jmarrec, line 227: New PathInputView class

jmarrec:
@macumber ping

src/shared_gui_components/MeasureManager.cpp
```diff
@@ -607,8 +607,10 @@ boost::optional<measure::OSArgument> MeasureManager::getArgument(const measure::
 
   } else if (type.value() == measure::OSArgumentType::Path) {
 
+    // TODO: aside from the fact that these arguments are weird / incorrectly named, neither the BCL-gem schema nor OS SDK actually handle them so
+    // they are not even inside the measure.xml
     bool isRead = argument.get("is_read", Json::Value(false)).asBool();
```
macumber, line 612: Does `isRead` mean that the file should be read from the path vs written to the path?

src/shared_gui_components/EditView.cpp
macumber: I think the important thing is to allow the user to ignore the file filter if they want to, so maybe always append "Any files (*)"?  The filters argument is a QStringList right?

src/shared_gui_components/EditView.cpp
```diff
@@ -400,4 +402,80 @@ void InputCheckBox::setIncomplete(bool incomplete) {
...
+void PathInputView::setDisplayValue(const QVariant& value) {
+  lineEdit->setText(value.toString());
+}
+
+void PathInputView::onSelectPathButtonClicked() {
+  QString lastPath = lineEdit->text();
+
+  QString selectedPath;
+
+  if (m_extension.isEmpty()) {
```
macumber, line 463: This logic seems reasonable, should we always append "Any files (*)" to the extension?  The docs for getSaveFileName say:

> Multiple filters are separated with ';;'

`.

src/shared_gui_components/EditView.cpp
```diff
@@ -400,4 +402,80 @@ void InputCheckBox::setIncomplete(bool incomplete) {
...
+  } else {
+    if (m_isRead) {
+      selectedPath = QFileDialog::getOpenFileName(nullptr, tr("Open Read File"), lastPath, m_extension);
+    } else {
+      selectedPath = QFileDialog::getSaveFileName(nullptr, tr("Select Save File"), lastPath, m_extension);
+    }
+  }
+
+  if (!selectedPath.isEmpty()) {
+    lineEdit->setText(selectedPath);  // QFileInfo(selectedPath).absoluteFilePath()
```
macumber, line 476: Will selectedPath always be absolute?  What if the user wants it to be relative?  What would it be relative to?

src/shared_gui_components/EditView.cpp
```diff
@@ -400,4 +402,80 @@ void InputCheckBox::setIncomplete(bool incomplete) {
...
+  : InputView(), lineEdit(new QLineEdit()), nameLabel(new QLabel()), m_isRead(isRead) {
+  auto* vLayout = new QVBoxLayout();
+  vLayout->setContentsMargins(0, 0, 0, 0);
+  vLayout->setSpacing(5);
+  setLayout(vLayout);
+
+  nameLabel->setOpenExternalLinks(true);
+  nameLabel->setWordWrap(true);
+  vLayout->addWidget(nameLabel);
+
```
macumber, line 415: Does the `lineEdit` need to be connected to `emit selectedPathChanged(toPath(selectedPath))`?  Does that support free text entry or is it read only?

src/shared_gui_components/EditView.cpp
macumber: No extension == directory is the only kind of fishy logic here, that should really be an attribute of the argument itself (`isDir` like `isRead`), if you are updating the SDK maybe this is worth adding?

macumber:
This is cool @jmarrec, I added some comments, maybe it is worth adding some new attributes if you are already updating the SDK?

src/shared_gui_components/MeasureManager.cpp
jmarrec: I think that was the intent. not sure

src/shared_gui_components/EditView.cpp
jmarrec: Yes, for sure need some work: https://github.com/NREL/OpenStudio/issues/5273

src/shared_gui_components/EditView.cpp
jmarrec: it's ";;" separated string.

Effectively, right now, until the SDK gets updated, the extension is always "All Files(*)"

jmarrec:
@macumber until #5273 is addressed, I'd say we just merge this.

src/shared_gui_components/EditController.cpp
```diff
@@ -251,6 +252,22 @@ InputController::InputController(EditController* editController, measure::OSArgu
...
+
+    if (m_argument.hasValue()) {
+      pathInputView->lineEdit->setText(QString::fromStdString(m_argument.valueAsString()));
+    } else if (m_argument.hasDefaultValue()) {
+      pathInputView->lineEdit->setText(QString::fromStdString(m_argument.defaultValueAsString()));
+    }
+
+    connect(pathInputView, &PathInputView::selectedPathChanged, this,
+            static_cast<void (InputController::*)(const openstudio::path&)>(&InputController::setValue));
+    connect(pathInputView->lineEdit, &QLineEdit::textEdited, [this](const QString& v) { this->setValue(toPath(v)); });
```
jmarrec, line 268: HEre the lineEdit is already connected

src/shared_gui_components/EditView.cpp
jmarrec: Already done, see https://github.com/openstudiocoalition/OpenStudioApplication/pull/761#pullrequestreview-2421146641

src/shared_gui_components/EditView.cpp
macumber: @jmarrec do you think the guidance should be that file path arguments are always absolute, used for reading/writing files outside of the workflow.  If there are files written by the measure, read by other measures, as part of the workflow those would just be found by file name (a string argument) using `WorkflowJSON::findFile` which searches the workflow paths?

PR #765, opened 2024-10-29
Fix #763 - Adjust deps for debian bookworm

### LINKED ISSUE #763: Support Debian
Enhancement Request
--------------
Ubuntu is a derivative of Debian. Why not supporting both?

## Detailed Description

When trying to install the `.deb` package in Debian (12, bookworm, kernel 6.1.0-26-amd64), I get the following message, quite similar to [#375 ](https://github.com/openstudiocoalition/OpenStudioApplication/issues/375):
```
$ sudo apt install ./OpenStudioApplication-1.8.0+2722e3e751-Ubuntu24.04.deb 
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Note, selecting 'openstudioapplication-1.8.0' instead of './OpenStudioApplication-1.8.0+2722e3e751-Ubuntu24.04.deb'
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 openstudioapplication-1.8.0 : Depends: libwxgtk3.0-0v5 (>= 3.0.0) but it is not installable or
                                        libwxgtk3.0-0 (>= 3.0.0) but it is not installable or
                                        libwxgtk3.0-gtk3-0v5 (>= 3.0.0) but it is not installable or
                                        libwxgtk3.2-1t64 (>= 3.2.0) but it is not installable
```

I know that the openstudiocoalition.org site explicitly says: _OpenStudio Application is supported on 64-bit versions of Windows, OS X, and Ubuntu_. It would be great if Debian was supported as well. Perhaps it is supposed to work in Debian as well. Is it?

jmarrec:
* Fix #763

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -863,7 +863,7 @@ elseif(UNIX)
   # set(CPACK_DEBIAN_PACKAGE_DEBUG ON)
 
   # Focal Fossa (20.04) uses libwxgtk3.0-gtk3-0v5, apparently need libxcb-xinerama0 as well
-  set(CPACK_DEBIAN_PACKAGE_DEPENDS "libxcb-xinerama0 (>= 1.13), libwxgtk3.0-0v5 (>= 3.0.0) | libwxgtk3.0-0 (>=3.0.0) | libwxgtk3.0-gtk3-0v5 (>=3.0.0) | libwxgtk3.2-1t64 (>= 3.2.0), libxcb-cursor0")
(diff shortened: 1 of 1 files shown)


jmarrec:
User reported on https://github.com/openstudiocoalition/OpenStudioApplication/issues/763#issuecomment-2446446588 that the 22.04 deb installer now works for Debian Bookworm. Maybe we should had that to the release notes or the wiki somewhere

PR #771, opened 2024-11-27
Update to OS 3.9.0

macumber:


.github/workflows/app_build.yml
```diff
@@ -71,8 +71,8 @@ jobs:
           COMPRESSED_EXT: tar.gz
           BINARY_PKG_PATH: _CPack_Packages/Darwin/IFW
           COMPRESSED_PKG_PATH: _CPack_Packages/Darwin/TGZ
-          MACOSX_DEPLOYMENT_TARGET: 11.0
-          SDKROOT: /Applications/Xcode_15.0.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
+          MACOSX_DEPLOYMENT_TARGET: 13.0
+          SDKROOT: /Applications/Xcode_15.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk
```
jmarrec, line 75: We should make sure to update the release notes (and maybe wiki), since this is a change that may affect older macs @macumber

PR #775, opened 2024-12-10
Update installer qss script

macumber:
Add back in method to click next on the select components page (leave all component selections to the defaults) and un-select run after install when using scripted installer.

### DIFF
--- a/ci/install_script_qtifw.qs
+++ b/ci/install_script_qtifw.qs
@@ -66,9 +66,9 @@ Controller.prototype.TargetDirectoryPageCallback = function()
 };
 
 // Unused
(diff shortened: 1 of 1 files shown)


jmarrec:
I'm sure you've tested the changes, and they're fine. But this is actually deprecated and should be totally removed, the workflow that use it replaced, and documentation update.

QtIFW 4.0.0 added a CLI feature. 

```
./OpenStudioApplication-x.y-z.exe --accept-licenses --default-answer --confirm-command --root ${installDir} install
```

https://doc.qt.io/qtinstallerframework/ifw-use-cases-cli.html#unattended-usage

jmarrec:
- Filed at #776

PR #780, opened 2025-01-03
Show empty required object-list fields with warning icon, remove unused IGdefaultRemoved slot

### LINKED ISSUE #778: Unable to assign zone name to Refrigeration:Case
Issue overview
--------------

Toggling the Zone Name drop down does not update the zone name for Refrigeration:Case object.
image
image

## Steps to Reproduce

Here is a sample .osm file: 
[file f1basic dr-badran.osm.txt](https://github.com/user-attachments/files/18265071/file.f1basic.dr-badran.osm.txt)

1.
2.
3.
4.

## Details
If another thermal zone is added to the project, one can toggle the drop down, select the other thermal zone, then toggle the drop down again and select the correct thermal zone to get OpenStudio to write to the .osm file.
### Environment

 - Platform (Operating system, version): Windows 11
image

macumber:
Fixes #778

### DIFF
--- a/src/model_editor/InspectorGadget.cpp
+++ b/src/model_editor/InspectorGadget.cpp
@@ -789,9 +789,12 @@ void InspectorGadget::layoutComboBox(QVBoxLayout* layout, QWidget* parent, opens
   int idx = combo->findText(curVal.c_str(), Qt::MatchFixedString);
 
   if (-1 == idx) {
+
+    // field is currently set to an invalid value
     idx = 0;
-    connect(combo, static_cast<void (QComboBox::*)(const QString&)>(&QComboBox::currentTextChanged), this, &InspectorGadget::IGdefaultRemoved);
-    //QString errormsg("We have a value:");
+    combo->insertItem(idx, QPixmap(":/images/alert_image.png"), curVal.c_str(), "Invalid");
+
+   //QString errormsg("We have a value:");
     //errormsg += curVal.c_str();
     //errormsg += " that does not match the allowable values in the idd.Name:";
     //errormsg += name.c_str();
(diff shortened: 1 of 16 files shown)


src/model_editor/InspectorGadget.cpp
```diff
@@ -789,9 +789,12 @@ void InspectorGadget::layoutComboBox(QVBoxLayout* layout, QWidget* parent, opens
   int idx = combo->findText(curVal.c_str(), Qt::MatchFixedString);
 
   if (-1 == idx) {
+
+    // field is currently set to an invalid value
     idx = 0;
-    connect(combo, static_cast<void (QComboBox::*)(const QString&)>(&QComboBox::currentTextChanged), this, &InspectorGadget::IGdefaultRemoved);
```
macumber, line 793: The `IGdefaultRemoved` slot was only connected in this case and it was being called directly from `InspectorGadget::layoutComboBox` at line 808 ` combo->setCurrentIndex(idx)`.  This means that the slot was called before the user had a chance to do anything.

jmarrec:
@macumber A before/after screenshot or gif would go a long way in helping to see the changes when reviewing.

jmarrec:
I'm not seeing any icons. Model with zero thermal zones:

image

One thermal zone: it defaults to "Thermal Zone 1"

image

jmarrec:
**After, this PR**

Model with no thermal zone:

image

macumber:
I saw the same issue and figured that behavior was probably better than whatever bugs I would introduce trying to clean up the combo box list when the user picked a new option.  I heard you on the gifs for PRs, I'll try to be better :-)

PR #793, opened 2025-02-27
Fix #792 - Add support for HotWaterEquipment and HotWaterEquipmentDefinition

### LINKED ISSUE #792: Add support for HotWaterEquipment and HotWaterEquipmentDefinition
Enhancement Request
--------------

* We do not have a HotWaterEquipmentDefinition in the library 
* If I add one to my model, I can drag and drop it on a SpaceType/Space, but it doesn't have an icon
* It is not visible on the Loads tab

## Detailed Description

Image

Not on the Loads tab

Image

## Possible Implementation

* Add an icon for the HotWaterEquipmentDefinition
* Add a section on the Loads tab

jmarrec:
* Fix #792


src/openstudio_lib/HotWaterEquipmentInspectorView.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  mainGridLayout->setRowStretch(8, 100);
+
+  label =
+    new QLabel("<i>The object models hot water equipment in the zone which consumes district heating, such as cooking equipment or process loads. "
+               "All of the energy consumed by the equipment becomes a heat gain in the zone or is lost (exhausted). "
+               "This object consumes district heating energy directly and does not cause a load on a hot water plant loop or water heater. "
+               "For domestic hot water uses, such as sinks and showers, see WaterUse:Equipment.</i>");
+  label->setWordWrap(true);
+  mainGridLayout->addWidget(label, 9, 0, 1, 4);
```
jmarrec, line 111: Add a footnote clarifying what the object is. There is no precedent, so I'd like people to weight in on whether this is helpful or not, anbd whether it should be displayed differently in the UI.

Made it look like this

image


@macumber @marccosta @chrisbalbach @craigers290  @antonszilasi

macumber: Yeah, I was looking at the IDD and I/O Reference trying to figure out why this wasn't added in the first place.  It is a bit confusing that it doesn't hook up to a water loop.  I think the text is helpful and looks pretty good.  If we wanted to be super fancy, we could add a "for more information" link to the I/O Reference.  That could be a general pattern but it might be too hard to maintain though.

jmarrec: It's really the dhw equivalent of the ZoneHVACIdealLoadsAirSystem.

I agree that I don't want to keep external links up to date.

Issue #166, opened 2015-04-15
Sort measures alphabetically in GUI


DavidGoldwasser:
@macumber @axelstudios maybe this is related to API improvements mentioned in issue NREL/OpenStudio#1609.

The comment on YouTube about not finding measures is more common of an issue now that we have enough measures that results often span multiple pages. Ideally If I have enough screen space I'd like to see more than 10 at a time, but if I see 10 at a time, they should be alphabetical. Currently they are reverse of order last edited.
https://www.youtube.com/watch?v=umQR8XoORZc 

![4-15-2015 9-32-54 am](https://cloud.githubusercontent.com/assets/4857055/7168019/89ea7d04-e377-11e4-867a-9208e1fce82f.png)

![4-15-2015 9-33-16 am](https://cloud.githubusercontent.com/assets/4857055/7168023/8eec104c-e377-11e4-8569-32ae42d447ff.png)

macumber:
Fixed in https://github.com/openstudiocoalition/OpenStudioApplication/pull/795

macumber:
Reverting this change in https://github.com/openstudiocoalition/OpenStudioApplication/pull/824 due to https://github.com/openstudiocoalition/OpenStudioApplication/issues/823

Opened https://github.com/BuildingComponentLibrary/bcl-manifest/issues/47 as a request to add sorting to the public BCL API

PR #795, opened 2025-03-10
Issue #166 - Sort measures alphabetically in GUI

antonszilasi:


src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -190,16 +217,17 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
     delete comp;
   }
 
-  RemoteBCL remoteBCL;
-  remoteBCL.setTimeOutSeconds(m_timeoutSeconds);
-  std::vector<BCLSearchResult> responses;
-  if (filterType == "components") {
-    responses = remoteBCL.searchComponentLibrary(searchString.toStdString(), tid, pageIdx);
-  } else if (filterType == "measures") {
-    responses = remoteBCL.searchMeasureLibrary(searchString.toStdString(), tid, pageIdx);
```
antonszilasi, line 199: It was too hard to refractor all the methods around this, instead I fetched all the responses from the BCL if it is the first page or the m_allRepsonses is empty (could just do the latter) then I sort them alphabetically

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -190,16 +217,17 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
...
+  if (pageIdx == 0 || m_allResponses.empty()) {
+    m_allResponses = fetchAndSortResponses(filterType, tid, searchString);
   }
 
-  for (const auto& response : responses) {
+  // Paginate responses
+  int itemsPerPage = 10;  // Assuming 10 items per page
```
antonszilasi, line 225: I think this is a far assumption? items per page should  be 10?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,6 +168,34 @@ void BuildingComponentDialogCentralWidget::setTid() {
   requestComponents(m_filterType, m_tid, m_pageIdx, m_searchString);
 }
 
+std::vector<openstudio::BCLSearchResult> BuildingComponentDialogCentralWidget::fetchAndSortResponses(const std::string& filterType, int tid, const QString& searchString) {
```
antonszilasi, line 171: will add tests for this shortly

antonszilasi:
Animation

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,6 +168,34 @@ void BuildingComponentDialogCentralWidget::setTid() {
...
+  m_allResponses.clear();
+  RemoteBCL remoteBCL;
```
macumber, line 173: You can set `RemoteBCL::setResultsPerQuery` to 100 which might make this go faster:

https://github.com/NREL/OpenStudio/blob/develop/src/utilities/bcl/RemoteBCL.cpp#L950

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,6 +168,34 @@ void BuildingComponentDialogCentralWidget::setTid() {
...
+    std::vector<BCLSearchResult> pageResponses;
+    if (filterType == "components") {
+      pageResponses = remoteBCL.searchComponentLibrary(searchString.toStdString(), tid, currentPage);
+    } else if (filterType == "measures") {
+      pageResponses = remoteBCL.searchMeasureLibrary(searchString.toStdString(), tid, currentPage);
+    }
+    responses.insert(responses.end(), pageResponses.begin(), pageResponses.end());
+    totalPages = remoteBCL.numResultPages();
+    currentPage++;
+  } while (currentPage < totalPages);
```
macumber, line 189: Just a `C++`ism

```suggestion
  } while (++currentPage < totalPages);
```

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -190,16 +217,17 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
...
   }
 
-  RemoteBCL remoteBCL;
-  remoteBCL.setTimeOutSeconds(m_timeoutSeconds);
-  std::vector<BCLSearchResult> responses;
-  if (filterType == "components") {
-    responses = remoteBCL.searchComponentLibrary(searchString.toStdString(), tid, pageIdx);
-  } else if (filterType == "measures") {
-    responses = remoteBCL.searchMeasureLibrary(searchString.toStdString(), tid, pageIdx);
+  if (pageIdx == 0 || m_allResponses.empty()) {
```
macumber, line 220: This would be a bug if you changed `tid` or `searchString` but then called again with `pageIdx=0`.  You need something like `m_lastSearchString =  filterType + tid + searchString`, then clear `m_allResponses` when that last string cache key changes

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -190,16 +217,17 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
...
+    m_allResponses = fetchAndSortResponses(filterType, tid, searchString);
   }
 
-  for (const auto& response : responses) {
+  // Paginate responses
+  int itemsPerPage = 10;  // Assuming 10 items per page
+  int startIdx = pageIdx * itemsPerPage;
+  int endIdx = std::min(startIdx + itemsPerPage, static_cast<int>(m_allResponses.size()));
+  std::vector<BCLSearchResult> paginatedResponses(m_allResponses.begin() + startIdx, m_allResponses.begin() + endIdx);
```
macumber, line 228: This seems like it could generate out of bounds array access -> crashes.

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -190,16 +217,17 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
...
-    responses = remoteBCL.searchMeasureLibrary(searchString.toStdString(), tid, pageIdx);
+  if (pageIdx == 0 || m_allResponses.empty()) {
+    m_allResponses = fetchAndSortResponses(filterType, tid, searchString);
   }
 
-  for (const auto& response : responses) {
+  // Paginate responses
+  int itemsPerPage = 10;  // Assuming 10 items per page
+  int startIdx = pageIdx * itemsPerPage;
+  int endIdx = std::min(startIdx + itemsPerPage, static_cast<int>(m_allResponses.size()));
```
macumber, line 227:
```suggestion
  size_t startIdx = pageIdx * itemsPerPage;
  size_t endIdx = std::min(startIdx + itemsPerPage, static_cast<int>(m_allResponses.size()));
```

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -210,11 +238,11 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
   m_collapsibleComponentList->setText(title);
 
   // the total number of results
-  int lastTotalResults = remoteBCL.lastTotalResults();
+  int lastTotalResults = m_allResponses.size();
   m_collapsibleComponentList->setNumResults(lastTotalResults);
```
macumber, line 242: What do `numResults` and `numPages` do?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -210,11 +238,11 @@ void BuildingComponentDialogCentralWidget::setTid(const std::string& filterType,
...
 
   // the number of pages of results
-  int numResultPages = remoteBCL.numResultPages();
+  int numResultPages = (lastTotalResults + itemsPerPage - 1) / itemsPerPage;
```
macumber, line 245: Is this the same as `(lastTotalResults / itemsPerPage) + 1`?

macumber: Also make a comment when using integer division

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
macumber: Use `size_t` instead of `int` when doing math with iterators

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: Yes, your suggestion is more straight forward though so I will use it

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: I have refractored the code to remove these variables

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: This should be addressed by the lines


```
  if (newKey != currentKey) {
    openstudio::RemoteBCL remoteBCL;
    m_allResponses = fetchAndSortResponses(filterType, tid, searchString, remoteBCL);
  }
```

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
macumber: Did you try this @antonszilasi ?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,53 +168,87 @@ void BuildingComponentDialogCentralWidget::setTid() {
...
                                                   const QString& searchString) {
 
-  if (m_tid != tid || m_searchString != searchString) {
-    m_collapsibleComponentList->firstPage();
-  }
+  std::string newKey = std::to_string(tid) + filterType + searchString.toStdString();
+  std::string currentKey = std::to_string(m_tid) + m_filterType + m_searchString.toStdString();
 
-  m_filterType = filterType;
+  m_collapsibleComponentList->firstPage();
```
macumber, line 206: Do you only want to force navigating to the first page if the key changes?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,53 +168,87 @@ void BuildingComponentDialogCentralWidget::setTid() {
...
+  
+  if (!m_allResponses.empty()) {
+    size_t startIdx = pageIdx * itemsPerPage;
+    size_t endIdx = std::min(startIdx + itemsPerPage, m_allResponses.size());
+    std::vector<BCLSearchResult> paginatedResponses(m_allResponses.begin() + startIdx, m_allResponses.begin() + endIdx);
+  
+    for (const auto& response : paginatedResponses) {
+      auto* component = new Component(response);
+  
+      // TODO replace with a componentList owned by m_collapsibleComponentList
```
macumber, line 233: What is this TODO about?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: Im not sure how to measure this?

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: This was in the original code, but it appears as added here

`m_componentList->addComponent(component);`

I thought of removing it but maybe we should just keep it as we are keeping 

`m_componentList->addComponent(component);`

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
antonszilasi: Yes it seems simplier to do that since that was the original functionality

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
macumber: If it works, then it's almost certainly faster to do say 2 queries with 100 results per query than 20 queries with 10 results per query.  I'd just set the value to 100, set a break point to check that it is actually using that value for number of results, and then test to make sure it still works.

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
macumber: Updated

src/shared_gui_components/BuildingComponentDialogCentralWidget.cpp
```diff
@@ -169,53 +168,90 @@ void BuildingComponentDialogCentralWidget::setTid() {
...
+    }
+    responses.insert(responses.end(), pageResponses.begin(), pageResponses.end());
+    totalPages = remoteBCL.numResultPages();
+  } while (++currentPage < totalPages);
+
+  if (!responses.empty()) {
+    std::sort(responses.begin(), responses.end(), [](const BCLSearchResult& a, const BCLSearchResult& b) {
+      return a.name() < b.name();
+    });
+  }
```
jmarrec, line 197: If I'm following correctly, you just factored a bit of code into a function, and added this bit that sorts the results.

---

Side note: the `if (!responses.empty())` is not technically needed and does not make it faster. https://quick-bench.com/q/5lPbGyQNKctVRq7IdoD_fUVTPt4

(Running it a bunch of times locally produces results where one is sometimes faster than the other, so they're the same).

```
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
SortNoIf_Items         4732 ns         4730 ns       147972
SortWithIf_Items       5040 ns         5038 ns       138705
```

PR #809, opened 2025-06-17
Update to OS SDK v3.10.0 and EnergyPlus v25.1.0-WithDSOASpaceListFixes

jmarrec:


jmarrec:
mac is failing because there's a new workaround file `cpprestsdk_char_traits_workaround.hpp` that isn't shipped in the install package.

PR #821, opened 2025-07-07
Fix #815 - Add OutputControl:ResilienceSummaries to OS App

### LINKED ISSUE #815: [OS SDK 3.10.0] Add OutputControl:ResilienceSummaries
Enhancement Request
--------------

* https://github.com/NREL/OpenStudio/pull/5365/

jmarrec:
* Fix  #815

### DIFF
--- a/src/openstudio_lib/SimSettingsView.cpp
+++ b/src/openstudio_lib/SimSettingsView.cpp
@@ -32,6 +32,8 @@
 #include <openstudio/model/OutputTableSummaryReports_Impl.hpp>
 #include <openstudio/model/OutputDiagnostics.hpp>
 #include <openstudio/model/OutputDiagnostics_Impl.hpp>
+#include <openstudio/model/OutputControlResilienceSummaries_Impl.hpp>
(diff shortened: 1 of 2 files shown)


macumber:
Looks good, seems like there is no reason to be able to delete the object, this will trigger an update when opening an OSM that doesn't have the object but that is not a big deal.  

The only real use I saw for this is changing the output of Output:Table:ReportPeriod, is that being added to the StandardReport measure?  Should we update https://github.com/openstudiocoalition/openstudio-coalition-measures/tree/main/models/ShoeboxExample/measures/openstudio_results which is where the default results measure comes from?  We should update that repo too (and add that to the steps when releasing a new version, I had forgotten about it...)

macumber (approved):

jmarrec:
I will affect the ThermalResilienceSummary too. There is no need to update openstudio_results.
I don't expect people to actually use that object much, but anyways it does NOT change the structure of the table, it just changes how the Zone Heat Index is calculated.

PR #837, opened 2025-09-01
Work In Progress - Replace QtCharts with LGPL compliant JKQtPlotter

### LINKED ISSUE #836: OpenStudio Application uses GPLv3 only licensed QtCharts
The OpenStudio Application 1.9.0 Help -> About window implies that all the Qt modules used by the OpenStudio Application are available under GPL or LGPL, but I don't think that's the case.
<br>
Image
<br>
One of the modules used in the code is QtCharts, which the Qt6.5 page for states it is only available under under Commercial or GPLv3. https://doc.qt.io/qt-6.5/qtcharts-index.html
<br>
Image
<br>
This would make the OpenStudio Application and all its code dual licensed under the OpenStudio license and the GNU GPLv3.

macumber:
Fixes #836

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -546,7 +546,7 @@ message(STATUS "QT_INSTALL_DIR=${QT_INSTALL_DIR}")
 
 find_package(Qt6 ${QT_VERSION} COMPONENTS CoreTools GuiTools WidgetsTools QmlTools WebEngineCoreTools REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
 
-find_package(Qt6 ${QT_VERSION} COMPONENTS Core Core5Compat Gui Widgets Sql Svg Network Xml Concurrent PrintSupport Quick QuickWidgets Qml WebChannel Positioning WebEngineCore WebEngineWidgets REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
+find_package(Qt6 ${QT_VERSION} COMPONENTS Core Core5Compat Gui Widgets Sql Svg Network Xml Concurrent OpenGL OpenGLWidgets Positioning PrintSupport Quick QuickWidgets Qml WebChannel WebEngineCore WebEngineWidgets REQUIRED PATHS ${QT_INSTALL_DIR} NO_DEFAULT_PATH)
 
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -658,10 +659,23 @@ if(UNIX)
...
+# Make JKQTPlotter available after finding Qt
 include(FetchContent)
 
+set(JKQtPlotter_BUILD_EXAMPLES OFF)
+
+FetchContent_Declare(JKQTPlotter
+                     GIT_REPOSITORY https://github.com/jkriege2/JKQtPlotter.git
+                     GIT_TAG        v4.0.3)
+
+FetchContent_MakeAvailable(JKQTPlotter)
```
macumber: @jmarrec we could make a custom conan recipe for this dependency if that is better, but I don't know how to do that

src/openstudio_app/CMakeLists.txt
```diff
@@ -188,7 +188,7 @@ if( APPLE )
 
 elseif( UNIX )
 
-  foreach(LIB ${QT_LIBS} ${QT_WEB_LIBS})
+  foreach(LIB ${QT_LIBS} ${QT_WEB_LIBS} ${JKQT_LIBS})
```
macumber: @jmarrec this is not working on Unix but I'm not sure why

macumber:
Seem to be hitting https://github.com/jkriege2/JKQtPlotter/issues/80 on Mac at least

CMakeLists.txt
jmarrec: * https://github.com/conan-io/conan-center-index/issues/29109

jmarrec: * https://github.com/conan-io/conan-center-index/issues/29109

src/openstudio_app/CMakeLists.txt
jmarrec: This isn't even built when you are configuring (`_deps/jkqtplotter6-build/output/` is empty).

Also, I don't think we want to be doing this anyways.

We should link to the JKQTPlotter::JKQTPlotter target.

And I think it handles qtdeploy... Or we should just build it statically.

macumber:
I'm good to abandon this for now if you want @jmarrec

macumber:
Given https://github.com/openstudiocoalition/OpenStudioApplication/pull/859, QtCharts is too nice to get rid of.  

This is addressed by simply listing the licenses in https://github.com/openstudiocoalition/OpenStudioApplication/pull/842

PR #842, opened 2025-10-02
Fix #836 - QtCharts is GPL only

### LINKED ISSUE #836: OpenStudio Application uses GPLv3 only licensed QtCharts

jmarrec:
* Fix #836 

image


jmarrec:
@macumber I'm going to drop this PR in for now, which doesn't prevent us from doing this
* #837

PR #845, opened 2025-12-08
Update to OpenStudio SDK v3.11.0-RC1

jmarrec:
* Use the https://github.com/NREL/OpenStudio/releases/tag/v3.11.0-rc1
* There's a known issue for Ubuntu https://github.com/NREL/OpenStudio/issues/5543
* There is no mac intel x64 installer available
* I had to remove the use of the `Singleton<>` template since that got removed on OS SDK as well. Replaced with a Meyers singleton.
    * cf https://github.com/NREL/OpenStudio/pull/5450


jmarrec:
macos-13 is now fully removed, but that's convenient to leave it like this since the OS SDK is missing x64 mac insallers anyways. Not going to update the workflow to macos-15-intel now

jmarrec:
Ok CI looks good! @macumber wanna do a quick review before I merge?

src/model_editor/Application.hpp
```diff
@@ -20,12 +18,17 @@ namespace openstudio {
 
 /** Singleton application wide configuration management.
   */
-class ApplicationSingleton
+class Application
```
macumber, line 21: May require SketchUp Plug-in changes.

FindOpenStudioSDK.cmake
```diff
@@ -17,37 +17,37 @@ set(OPENSTUDIO_EXT "tar.gz")
 if(APPLE)
   set(OPENSTUDIO_PLATFORM "Darwin-${ARCH}")
   if(ARCH MATCHES "arm64")
-    set(OPENSTUDIO_EXPECTED_HASH b412f1cd6ab6747c75dece567291e81e)
+    set(OPENSTUDIO_EXPECTED_HASH 389a02d7f9d3526454bca74393ac7a6e)
   else()
-    set(OPENSTUDIO_EXPECTED_HASH 00d0321aa6c4ab5a3c36fe2335a07157)
+    set(OPENSTUDIO_EXPECTED_HASH TODO_MISSING_TODO) # There is no mac intel (x86_64) installer
```
macumber, line 22: Will there be a Mac x86 installer or is this the end for that platform?

macumber (approved):

FindOpenStudioSDK.cmake
jmarrec: I am hoping there will be one, but not sure.

PR #846, opened 2025-12-22
Update to MacOS 15 runners

macumber:
ARM based package is working, the Intel one is failing due to missing x86 build of OpenStudio


conan.lock
```diff
@@ -4,7 +4,7 @@
         "zlib/1.3.1#af8f2ef3d8e0fddf1975281d7fe43bf6%1705999193.687",
         "websocketpp/0.8.2#842a0419153a8aa52f3ea3a1da557d38%1695972005.713",
         "tinygltf/2.5.0#65c28d0a4c3cbd4ef92b08b59df769da%1701621757.442",
-        "termcap/1.3.1#1986f84bf21dd07ea774b027a3201fcb%1678542508.75",
+        "termcap/1.3.1#479400c750a869f77b3d2d3a82e06f7d%1755703795.492",
```
macumber, line 7: There is a newer version of this package which raises the minimum version of CMake.  However, the OpenStudio conan.lock pins the older revision which doesn't work with the version of CMake installer on the new Mac runners

PR #856, opened 2026-02-05
Future proof pinging bcl.nlr.gov instead of bcl.nrel.gov

jmarrec:


src/shared_gui_components/NetworkProxyDialog.cpp
```diff
@@ -131,7 +131,10 @@ std::pair<QNetworkReply::NetworkError, QString> NetworkProxyDialog::testProxyCon
   QNetworkReply* head = nam.head(QNetworkRequest(QUrl("https://bcl.nrel.gov/")));
 
   if (!head) {
-    return std::make_pair(QNetworkReply::UnknownNetworkError, QString("Unknown error creating connection to proxy."));
+    QNetworkReply* head = nam.head(QNetworkRequest(QUrl("https://bcl.nlr.gov/")));
+    if (!head) {
+      return std::make_pair(QNetworkReply::UnknownNetworkError, QString("Unknown error creating connection to proxy."));
+    }
```
jmarrec, line 137: I think bcl.nrel.gov will be gone soon, but right now bcl.nlr.gov doesn't exist yet, so just test both.

This only happens when testing proxy connection, which I guess not a lot of people use anyways.

macumber: Yeah this is a good idea.  Ideally, we could make the BCL URL configurable and guess a smart default for the whole app like this.  However, it looks like there isn't really a good way to configure the URL in https://github.com/NatLabRockies/OpenStudio/blob/develop/src/utilities/bcl/RemoteBCL.cpp.

jmarrec: Yes, the OS SDK is what really queries the BCL. If bcl.nrel.gov were to disappear, every known version of the OS SDK until 3.11.0 included would no longer be able to connect to the BCL, so I do hope there's going to be a DNS redirection for the foreseeable future.

macumber:
@jmarrec do you know when this new url would be live? It still isn't up for me

jmarrec:
@DavidGoldwasser Do know when the url for the BCL will change please?

jmarrec:
Ok it's effective, https://bcl.nrel.gov/ is no longer, https://bcl.nlr.gov/ is

jmarrec:
@macumber should we merge?

PR #859, opened 2026-03-03
#757 - Implement a Site subtab for ground temperatures

### LINKED ISSUE #757: Implement a method to specify monthly ground temperature schedule
Enhancement Request
--------------
Implement a method to specify monthly ground temperature schedule.

## Detailed Description
Currently there seems to be no way in the GUI to specify a ground temperature schedule. As per information available online, this requires the use of a measure. Since in warm climates the default 18 degC ground temperature value (is it the case ?) is causing considerable errors in zone heat load, a feature to specify this (at least a constant value through out the year) is essential.

## Possible Implementation
1. Somewhere under `Thermal Zones`

jmarrec:
* Fix #757 

For now I added:

* Site:GroundTemperature:BuildingSurface
* Site:GroundTemperature:Shallow, with an option to import if the WeatherFile has Ground Temps at 0.5m
* Site:GroundTemperature:Deep, with an option to import if the WeatherFile has a Ground Temps at 4.0m
* Site:GroundTemperature:FCfactorMethod, with an option to import if the WeatherFile has Ground Temps at 0.5m
* Site:WaterMainsTemperature

Old screen [2026-03-05-757-GroundTemp](https://github.com/user-attachments/assets/f1d77bff-f538-4214-aada-07c3622f1933)

Latest work, where I added a Bar chart and the Site:GroundTemperature:FCfactorMethod

2026-03-05-bis-757-GroundTemp

# TODO:

* [x] While recording the inital GIF [2026-03-03-757-GroundTemp](https://github.com/user-attachments/assets/5b708528-5d4d-4f50-81b3-419183aac723) I'm noticing that I have a hiccup with the units being displayed when I switch back and forth between subtabs...
* I'm considering adding Kiva, but this is far more complicated
* [x] Ideally a SiteWaterMainsTemperature panel would be added too, this is a widely useful object as soon as you intend to model DHW


src/openstudio_lib/SiteWaterMainsTemperatureWidget.cpp
```diff
@@ -0,0 +1,224 @@
...
+// SiteWaterMainsTemperatureWidget
+
+SiteWaterMainsTemperatureWidget::SiteWaterMainsTemperatureWidget(bool isIP, QWidget* parent) : QWidget(parent), m_isIP(isIP) {
+  auto* mainLayout = new QVBoxLayout();
+  mainLayout->setAlignment(Qt::AlignLeft | Qt::AlignTop);
+  mainLayout->setContentsMargins(10, 10, 10, 10);
+  mainLayout->setSpacing(20);
+  setLayout(mainLayout);
+
+  auto* titleLabel = new QLabel("Site:WaterMainsTemperature");
```
macumber, line 81: You can delete this with the red X in the object type selector, but it isn't completely obvious.  It might be nice to add a delete button here too?

macumber:
This is really impressive @jmarrec, left one optional idea to add a second remove button inside the object inspector

macumber (approved):

src/openstudio_lib/SiteWaterMainsTemperatureWidget.cpp
jmarrec: I don't think we have a precedent for this, all other tabs work with the X in the object selector.

For now I'm not going to add it, if it still bothers you, we can do it.

PR #863, opened 2026-03-06
Implement ability to change model from the 3D Geometry Preview View (set Construction, ThermalZone, reverse Vertices)

### LINKED ISSUE #862: Implement ability to change model from the 3D Geometry Preview View (set Construction, ThermalZone, reverse Vertices)
Enhancement Request
--------------

It'd be awesome if we could stuff from the 3D Geometry Preview view like assigning thermal zones, space types to spaces, constructions and boundary conditions for surfaces, etc.

## Detailed Description

Would need to figure a way to go from JS to C++

## Possible Implementation

QWebChannel

jmarrec:
* Fix #862 

I am pretty darn excited about this. I think this is huge.

2026-03-06-Interative_geometry_preview

### DIFF
--- a/src/openstudio_lib/GeometryPreviewView.cpp
+++ b/src/openstudio_lib/GeometryPreviewView.cpp
@@ -11,7 +11,11 @@
 #include "../model_editor/Application.hpp"
 
 #include <openstudio/model/Model_Impl.hpp>
+#include <openstudio/model/PlanarSurface.hpp>
+#include <openstudio/model/PlanarSurface_Impl.hpp>
 #include <openstudio/model/ThreeJSForwardTranslator.hpp>
+
+#include <algorithm>
 
 #include <openstudio/utilities/core/Assert.hpp>
 #include <openstudio/utilities/idd/IddEnums.hxx>
@@ -22,12 +26,24 @@
 #include <QPushButton>
 #include <QFile>
 #include <QWebEngineScriptCollection>
+#include <QWebChannel>
 #include <QtConcurrent>
 #include <QCheckBox>
 
 using namespace std::placeholders;
 
 namespace openstudio {
+
+GeometryBridge::GeometryBridge(model::Model& model, QObject* parent) : QObject(parent), m_model(model) {}
+
+void GeometryBridge::reverseSurfaceVertices(const QString& surfaceName) {
+  if (auto surface = m_model.getModelObjectByName<model::PlanarSurface>(surfaceName.toStdString())) {
+    auto vertices = surface->vertices();
+    std::reverse(vertices.begin(), vertices.end());
+    surface->setVertices(vertices);
+    emit modelChanged();
+  }
+}
 
 GeometryPreviewView::GeometryPreviewView(bool isIP, const openstudio::model::Model& model, QWidget* parent) : QWidget(parent) {
   // TODO: DLM implement units switching
@@ -76,6 +92,15 @@ PreviewWebView::PreviewWebView(bool isIP, const model::Model& model, QWidget* t_
   m_view = new QWebEngineView(this);
   m_page = new OSWebEnginePage(m_view);
   m_view->setPage(m_page);  // note, view does not take ownership of page
+
+  auto* channel = new QWebChannel(m_page);
+  m_bridge = new GeometryBridge(m_model, this);
+  channel->registerObject(QStringLiteral("bridge"), m_bridge);
+  m_page->setWebChannel(channel);
+  connect(m_bridge, &GeometryBridge::modelChanged, this, [this]() {
+    m_json = QString();
+    refreshClicked();
+  });
 
   auto* mainWindow = OSAppBase::instance()->currentDocument()->mainWindow();
   const bool verboseOutput = mainWindow->geometryDiagnostics();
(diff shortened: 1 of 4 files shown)


eringold:
this is awesome. can you do a multi-select?

jmarrec:
Dude! let me be content for a day before asking for more 🙃 

(I think that'd require a significant change)

src/openstudio_lib/GeometryPreviewView.cpp
```diff
@@ -11,24 +11,149 @@
...
+void GeometryBridge::reverseSurfaceVertices(const QString& surfaceName) {
+  if (auto surface = m_model.getModelObjectByName<model::PlanarSurface>(surfaceName.toStdString())) {
+    auto vertices = surface->vertices();
+    std::reverse(vertices.begin(), vertices.end());
+    surface->setVertices(vertices);
+    emit modelChanged();
+  }
+}
+
+void GeometryBridge::triangulateSurface(const QString& surfaceName) {
```
macumber, line 63: Just curious, do you have a use case for this?

src/openstudio_lib/GeometryPreviewView.cpp
```diff
@@ -11,24 +11,149 @@
...
+}
+
+void GeometryBridge::setWindExposure(const QString& surfaceName, const QString& value) {
+  if (auto surface = m_model.getModelObjectByName<model::Surface>(surfaceName.toStdString())) {
+    surface->setWindExposure(value.toStdString());
+    emit modelChanged();
+  }
+}
+
+void GeometryBridge::setOutsideBoundaryCondition(const QString& surfaceName, const QString& value) {
```
macumber, line 106: Setting the adjacent surface could be cool, although I guess you can do that from the surface tab too.

src/openstudio_lib/library/geometry_preview.html
```diff
@@ -868,6 +1056,309 @@
...
+      bridge.triangulateSurface(surfaceName);
+      contextMenu.style.display = 'none';
+    });
+  }
+  return btn;
+}
+
+function displaySelectedObjectContextMenu() {
+
+  contextMenu.innerHTML = '<h2>Actions</h2>';
```
macumber, line 1292: Is there a way to give this menu a close button?

src/openstudio_lib/GeometryPreviewView.cpp
```diff
@@ -11,24 +11,149 @@
...
 #include <QtConcurrent>
 #include <QCheckBox>
 
 using namespace std::placeholders;
 
 namespace openstudio {
 
+GeometryBridge::GeometryBridge(model::Model& model, QObject* parent) : QObject(parent), m_model(model) {}
+
+void GeometryBridge::reverseSurfaceVertices(const QString& surfaceName) {
```
macumber, line 54: Set surface type could be useful as well.

src/openstudio_lib/library/geometry_preview.html
macumber, line 1: Do you think it would be worthwhile to add DisplayName and CadObjectId to the info messages if they exist?

macumber:
This is really cool @jmarrec great job!

macumber:
Left some optional comments but approving as is.

macumber (approved):

src/openstudio_lib/GeometryPreviewView.cpp
jmarrec: For sun-related things like shading, having a non convex surfaces (eg a L, H shaped roof) is a problem, so I could see myself wanting to do this. I don't know, I was just in the flow and thought about adding it so I just did.

src/openstudio_lib/library/geometry_preview.html
jmarrec: Yeah, but is that necessary? Just click outside of it and it goes away. We don't have a close button on the left-click info panel (displaySelectedObjectDetails)

src/openstudio_lib/GeometryPreviewView.cpp
jmarrec: I think that's a valid point. 

You can also set the outside boundary condition and all from the surface tab too, it's just more tedious especially if you have auto-named surfaces like "Surface 114", which makes it quite hard to remember what you're touching.

src/openstudio_lib/GeometryPreviewView.cpp
jmarrec: I think that one definitely makes sense.

src/openstudio_lib/library/geometry_preview.html
jmarrec: Not 100% sure if that's valuable for people or not. I'm _guessing_ if you care about these, you're definitely using another 3D editor as the driver (eg: Rhino with Ladybug tools) and probably less likely to be trying to do that in the 3D GeometryPreview.

The question is whether this info is in the ThreeJSForwardTranslator export or not (I will have to check, cant remember)

It's a simple code change though for us, and most models won't have it, so I guess it would be fine to add it.

jmarrec: After checking, AFAIK this is not exported by the OpenStudio SDK. https://github.com/NatLabRockies/OpenStudio/blob/8008ef767fdc0f9d3dd3fabd383da15d009aef76/src/utilities/geometry/ThreeJS.hpp#L169

PR #864, opened 2026-03-15
Remove Qt from ModelEditor Ruby bindings

macumber:


jmarrec:
A couple of questions here: 

* SketchupInstallName.rb https://github.com/openstudiocoalition/OpenStudioApplication/blob/3507fb3482be58ec07dc2f3fa9e0df8dad6f1226/SketchUpInstallName.rb#L29 is that file still needed at all? If so, we should update it
    * Related: https://github.com/openstudiocoalition/openstudio-sketchup-plugin/issues/59 
* Is  QtCore5Compat still needed? I think we only use QRegExp from it, and we can replace with QRegularExpression

jmarrec:
* https://github.com/openstudiocoalition/OpenStudioApplication/pull/867

macumber:
Good to merge this @jmarrec ?

PR #865, opened 2026-03-17
Refactor with architecture and developer documentation

macumber:
# Intent

This PR started off with the goal to add architecture documentation for humans and llms to reference.  Instructions were added to prompt llms to review documentation for accuracy and suggest updates in future changes.  After reviewing the architecture, I felt compelled to clean it up.  I probably should have waited to do that on subsequent PRs but I continued on here to:

- Remove dead code: with separation of the OS Application and SketchUp Plug-in, the SWIG bindings to OS Application widgets are no longer needed.  These are not used by any other developer or application that I know of.  InspectorDialog and several of its widgets were only used in the SketchUp Plug-in and not needed in the OS Application. The *API headers were needed for linking shared libraries on Windows.  However, all libraries in the OS Application are compiled as static libraries. There were also some stray Jenkins based CI actions that are no longer needed. Removing all this dead code simplifies the build system and makes it more maintainable.  
- Architectural restructuring: shared_gui_components was not a real library. Its .cpp files were listed directly in openstudio_lib's CMakeLists.txt as sources, meaning they were compiled into openstudio_lib but physically lived in a separate directory for no particular reason.  There was no enforced boundary, so any file in the project could include anything else, and over time files that should have been small, reusable widgets were not due to dependencies on application-level code.  I wanted to define a proper library structure with rational libraries with correct dependency ordering. This required factoring out a new openstudio_qt_utils library for base classes and utilities that were needed across higher level libraries. A next step would be to break up openstudio_lib into smaller domain specific libraries but I did at least catch myself before doing that. I also want to remove BIMServer as well, but again have at least some common sense to not do that in this PR.
- Mechancal/style updates: now that we have a nice architectural separation of libraries I wanted to show it off by organizing include paths to be in order based on the libraries used.  A lot of includes had to be adjusted anyway due to files moving around so I figured we should reorder includes while were making those changes anyway.

# Changes

**Architecture / build system**

- New openstudio_qt_utils static library: extracts Application, OSProgressBar, QMetaTypes, PathWatcher, Utilities from model_editor
- New openstudio_shared_gui static library: separates shared_gui_components from openstudio_lib, eliminating the upward dependency
- QRC resources split between static libraries.
- OSItemId moved to shared_gui_components - the concrete seam that breaks the upward reference
- Strict link order enforced: OpenStudioApp -> openstudio_lib -> openstudio_shared_gui -> openstudio_qt_utils -> openstudioapp_utilities
- Removed Jenkinsfiles; CI fully on GitHub Actions
- Qt 6.11 locked as required version; Qt6QmlMeta and Qt6QmlWorkerScript promoted to REQUIRED

**Dead code removed**

- InspectorDialog class and test deleted (~1100 lines)
- TreeView, ModelObjectTreeWidget, ModelEditor.rc/.xml deleted
- Removed SWIG binding files (BIMServer.i, ModelEditor.i, Qt.i) and all *API.hpp export-macro headers which are not needed for a static build

**Developer documentation added**

- architecture.md - C4 diagrams, dependency graph, patterns, build guide, CI overview
- Per-library reference docs for all 7 targets
- llms.txt, PR review instructions, code refactor instructions, and doc-update prompt template

This is a large PR, most of the changes are file moves and small changes.  The main goal was to clean up the dependency graph:
image


macumber:
@jmarrec this is a crazy big PR but most of it is just due to moving files into a better dependency graph.  I have looked through all the files, I'm not sure if you want to or not.  If this is too big I could try to break it up somehow, possibly with https://github.github.com/gh-stack/.  I signed up for a preview of that, will see if we get access.

.github/workflows/app_build.yml
```diff
@@ -1,3 +1,21 @@
...
+# Caching: three independent caches keyed by OS + Conan profile + CACHE_KEY secret:
+#   1. ccache (compiler output cache)
+#   2. Conan package cache
+#   3. Qt + OpenStudio SDK download cache
+# Rotate the CACHE_KEY secret to bust all caches immediately.
+#
+# Required secrets: MACOS_DEVELOPER_ID_APPLICATION_CERTIFICATE_P12_BASE64,
+#   MACOS_DEVELOPER_ID_INSTALLER_CERTIFICATE_P12_PASSWORD, MACOS_KEYCHAIN_PASSWORD,
+#   NOTARIZATION_API_KEY, NOTARIZATION_API_TEAM_ID, NOTARIZATION_API_ISSUER_ID,
+#   SIGNPATH_CI_TOKEN, ANALYTICS_API_SECRET, ANALYTICS_MEASUREMENT_ID, CACHE_KEY
```
jmarrec, line 18: This is true right now, but will likely fall out of sync. Not sure what value it really provides. You can read the workflow to understand it. 

Should the workflow fail for some reason, you'll definitely need to understand it anyways

src/bimserver/CMakeLists.txt
```diff
@@ -18,26 +16,22 @@ set(${target_name}_test_src
...
-  openstudio_modeleditor
-  openstudio::openstudiolib
-  ${QT_LIBS}
+  openstudio_qt_utils
 )
 
-add_library(${target_name} ${${target_name}_src})
-target_link_libraries(${target_name} ${${target_name}_depends})
+add_library(${target_name} STATIC ${${target_name}_src} ${${target_name}_moc_src})
+target_link_libraries(${target_name} PUBLIC ${${target_name}_depends})
```
jmarrec, line 32: I don't really get why. We've already building a static library

CMake does not define the [BUILD_SHARED_LIBS](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html#variable:BUILD_SHARED_LIBS) variable by default, meaning without project or user intervention [add_library()](https://cmake.org/cmake/help/latest/command/add_library.html#command:add_library) will produce STATIC libraries.

src/bimserver/CMakeLists.txt
```diff
@@ -18,26 +16,22 @@ set(${target_name}_test_src
...
 AddPCH(${target_name})
 
-if(BUILD_MICRO_PROFILING)
-  target_link_libraries(${target_name} ${MICRO_PROFILER_LIB} )
-endif()
```
jmarrec, line 39: I'm guessing now that you explicitly link stuff in a flat/serial manner you are only linking OpenStudioApp (or perhaps openstudio_lib) to the micro profiler?

src/openstudio_lib/OSDocument.hpp
```diff
@@ -54,81 +61,55 @@ class OPENSTUDIO_API OSDocument : public OSQObjectController
...
+  boost::optional<IddObjectType> getIddObjectType(const OSItemId& itemId) const override;
+  boost::optional<model::ModelObject> getModelObject(const OSItemId& itemId) const override;
+  boost::optional<model::Component> getComponent(const OSItemId& itemId) const override;
+
+  // -------------------------------------------------------------------------
+  // OSDocument-specific API
+  // -------------------------------------------------------------------------
+
+  // TODO: promote to BaseDocument once MainWindow is extracted into an IMainWindow interface
+  // that lives in shared_gui_components or openstudio_qt_utils (no openstudio_lib dependency).
```
jmarrec: Does the TODO need to be addressed?

src/openstudio_lib/OSDocument.hpp
```diff
@@ -54,81 +61,55 @@ class OPENSTUDIO_API OSDocument : public OSQObjectController
...
-  // Returns the main window associated with this document.
+  // -------------------------------------------------------------------------
+  // BaseDocument interface — pure virtual overrides
+  // -------------------------------------------------------------------------
+
+  model::Model model() override;
+  bool modified() const override;
+  QString savePath() const override;
+  QString modelTempDir() const override;
+  model::Model componentLibrary() const override;
```
jmarrec, line 78: We lost the comments / docstrings here?

src/openstudio_lib/OSDocument.hpp
```diff
@@ -54,81 +61,55 @@ class OPENSTUDIO_API OSDocument : public OSQObjectController
...
-
-  // Returns the string path to the location where the document is saved.
-  // If the document is unsaved an empty string will be returned.
-  QString savePath() const;
-
-  // Returns the path to the directory where model resources are stored
-  QString modelTempDir() const;
-
-  // Returns the component library associated with this document.
-  openstudio::model::Model componentLibrary() const;
```
jmarrec, line 86: here

src/shared_gui_components/OSGridView.hpp
```diff
@@ -32,6 +32,13 @@ class OSItem;
 class GridCellLocation;
 class GridCellInfo;
 
+/**
+ * OSGridView is the QWidget that hosts and renders an OSGridController. It manages a QScrollArea
+ * containing a grid of OSCellWrapper widgets arranged in rows and columns. It handles row
+ * selection highlighting and forwards keyboard navigation events to the controller. Domain views
+ * (ThermalZonesGridView, SpaceTypesGridView, etc.) typically subclass OSGridView or compose it
+ * with filter/search bars.
+ */
```
jmarrec, line 41: Ok, that's actually probably helpful

src/shared_gui_components/RenderingColorWidget2.cpp
```diff
@@ -213,94 +212,4 @@ void RenderingColorWidget2::renderColorButtonClicked() {
   }
 }
 
-RenderingColorWidget::RenderingColorWidget(QWidget* parent) : QWidget(parent) {
```
jmarrec, line 216: I guess the idea is that RenderingColorWidget2 is kept, and RenderingColorWidget was unused and deleted?

jmarrec (commented):
This is obviously hard to review given how big it is

before I delve deeper and waste more than a few hours, @macumber can you please explain what you're trying to solve, why this is better, and do a walkthrough review of your changes (don't need to annotate everything, just one of each major change types).

.github/workflows/app_build.yml
macumber: I agree that manually maintaining this documentation would quickly become stale.  However, my thought was by adding the instruction "CI script added or modified (`ci/*.sh`, `ci/*.py`, `ci/*.qs`) → header comment must accurately describe usage, arguments, and exit codes." in `.github\instructions\pr-review.instructions.md` that we could rely on the co-pilot reviews to ensure it stays up to date.  

Part of the intent of this PR is to add documentation to guide llms in understanding the code and being able to make changes.  So, my thought was we could add documentation for llms and then rely on llms to help keep it up to date.  I'm open to removing it though.

src/bimserver/CMakeLists.txt
macumber: This change was just to make the static libs more explicit.  As part of dead code cleanup, I got rid of the *API header files which were only there in case we wanted to build shared libs.  I wanted to remove all of that unused code to make the project simpler and more maintainable.

src/bimserver/CMakeLists.txt
macumber: Yeah, we only have it linked in at the lowest `openstudioapp_utilities` level, it is then available to everything else in the project

src/openstudio_lib/OSDocument.hpp
macumber: Updated to this clearer TODO, I figured this PR was too big that I should not add this in now:

  // TODO: create a BaseMainWindow interface in shared_gui_components and change this 
  // to a virtual method `BaseMainWindow* mainWindow()` in BaseDocument.
  // This will allow us to split up openstudio_lib into domain-specific libraries
  // (e.g. constructions_gui, hvac_gui, etc.) which need to call mainWindow().

src/openstudio_lib/OSDocument.hpp
macumber: Added back

src/shared_gui_components/RenderingColorWidget2.cpp
macumber: Yeah RenderingColorWidget was no longer used

macumber:
> This is obviously hard to review given how big it is
> 
> before I delve deeper and waste more than a few hours, @macumber can you please explain what you're trying to solve, why this is better, and do a walkthrough review of your changes (don't need to annotate everything, just one of each major change types).

Thanks @jmarrec sorry this did get pretty big, I updated the PR description at the top to describe the intent of the changes

macumber:
@jmarrec it looks like the conan.openstudio.net certificate expired on 4/29/26:

Unable to connect to remote nrel-v2=https://conan.openstudio.net/artifactory/api/conan/conan-v2
1. Make sure the remote is reachable or,
2. Disable it with 'conan remote disable <remote>' or,
3. Use the '-nr/--no-remote' argument
Then try again.. Required by ''

anchapin:
@tijcolem said that he updated the cert. Let me know if it still has issues.

jmarrec:
@macumber If you feel confident, merge away.

macumber:
I will merge this after https://github.com/openstudiocoalition/OpenStudioApplication/pull/873 so I don't hold @Ski90Moo up

PR #876, opened 2026-05-24
feat: add clickable EnergyPlus doc links throughout OpenStudioApp UI

### LINKED ISSUE #160: Add contextual help for fields in OS App
Problem: Currently there is no direct way to get a description of an input field. You have to pull up the E+ Input/Output guide (you have to download it from energyplus.net since as of 8.3 it doesn’t ship with E+ anymore or look online)

Rationale: Adding contextual help would greatly help in having a streamlined workflow, staying in the OS App as much as possible.

Solution: Add contextual help directly in the GUI (hovering on the field, or a button, or something) or something that pulls the relevant Input/Output reference section in a PDF or online

Context: For example, if looking at a Chiller:Electric:EIR, there are a lot of fields. Let’s pick one at random: “Minimum Part Load Ratio”. It would be nice to be able to see either directly in the GUI (or have access to an online documentation at the proper section) what the Input/Output reference has to say about this input field:
“This numeric field contains the chiller’s minimum part-load ratio. The expected range is between 0 and 1. Below this part-load ratio, the compressor cycles on and off to meet the cooling load. The Minimum Part Load Ratio must be less than or equal to the Maximum Part Load Ratio. The default value is 0.1.”

Migrated from [UserVoice](http://openstudio.uservoice.com/forums/308629/suggestions/9592029)

Ski90Moo:
## Summary

Implements the enhancement requested in #160 — adds clickable hyperlinks from UI section headers and IDD type headers to the relevant EnergyPlus 25.1 Input/Output Reference pages on BigLadder Software. Tooltips show the full URL before clicking.

### Inspector / IDD object headers
- **`src/model_editor/IddObjectDocUrl.hpp`** (new): header-only hash map (`QHash<QString, QString>`) mapping 300+ `OS:*` IDD type names to their BigLadder EnergyPlus 25.1 I/O Reference URLs
- **`src/model_editor/InspectorGadget.cpp`**: renders the locked IDD type header in the right-sidebar inspector as a `Qt::RichText` hyperlink when a URL mapping exists

### Left-sidebar category headers
- **`src/openstudio_lib/OSCollapsibleItemHeader.hpp/.cpp`**: extends `OSCollapsibleItemHeader` with an optional URL; when set, the category name renders as a clickable blue link
- **`src/openstudio_lib/ModelObjectTypeListView.hpp/.cpp`**: adds a new constructor accepting `vector<tuple<IddObjectType, string, QString>>` to thread URLs through to each header
- **Loads tab**: People, Lights, Electric/Gas/Steam/Other Equipment, Internal Mass, Water Use Equipment, Hot Water Equipment definitions
- **Constructions subtab**: Constructions, Air Boundary, Internal Source, C-factor Underground Wall, F-factor Ground Floor
- **Materials subtab**: all 14 material types (standard, no-mass, air gap, all window material variants)
- **Other Schedules subtab**: Schedule Constant, Schedule Compact, Schedule File

### Section headers (SimSettings pattern)
- **`src/openstudio_lib/CollapsibleInspector.hpp/.cpp`**: extends `CollapsibleInspectorHeader` with an optional URL; section title renders as a clickable link
- **`src/openstudio_lib/SimSettingsView.cpp`**: 15 `CollapsibleInspector` section headers and 3 H1 labels (Run Period, Sizing Parameters, Timestep)
- **`src/openstudio_lib/LocationTabView.cpp`**: Weather File and Design Days headers
- **`src/openstudio_lib/YearSettingsWidget.cpp`**: Daylight Savings Time header
- **`src/openstudio_lib/GroundTemperatureView.hpp/.cpp`**: all 5 Ground Temperatures sidebar entries
- **`src/openstudio_lib/LifeCycleCostsTabView.cpp`**: Life Cycle Cost Parameters and NIST Fuel Escalation Rates headers
- **`src/openstudio_lib/BuildingInspectorView.cpp`**: North Axis label (Facility tab)

### Build infrastructure
- **`cmake/FindOpenStudioSDK.cmake`**: defines `BIGLADDERSOFTWARE_DOC_BASE_URL` as a single versioned CMake variable (EnergyPlus 25.1)
- **`src/model_editor/CMakeLists.txt`** and **`src/openstudio_lib/CMakeLists.txt`**: pass the URL as a compile definition so all files share a single source of truth for the base URL

## Test plan

- [ ] Open a model — confirm IDD type headers in the right-sidebar inspector render as blue hyperlinks and open the correct BigLadder page
- [ ] Navigate to Simulation Settings — confirm all CollapsibleInspector section headers and H1 labels are clickable links
- [ ] Check Loads, Constructions, Materials, and Other Schedules sidebar categories render as blue links
- [ ] Check Site tab (Weather File & Design Days, Ground Temperatures, Life Cycle Costs) section headers are linked
- [ ] Check Facility tab North Axis label is linked
- [ ] Hover over any link to confirm the tooltip shows the full URL before clicking
- [ ] Confirm object types with no URL mapping still display plain text (no broken link rendering)

Closes #160

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/model_editor/IddObjectDocUrl.hpp
```diff
@@ -0,0 +1,418 @@
...
+#ifndef MODELEDITOR_IDDOBJECTDOCURL_HPP
+#define MODELEDITOR_IDDOBJECTDOCURL_HPP
+
+#include <QString>
+#include <QHash>
+
+// Returns the BigLadder EnergyPlus 25.1 Input/Output Reference URL for the given
+// OS IDD type name (e.g. "OS:ThermalZone"), or an empty string if none is known.
+inline QString iddObjectDocUrl(const QString& iddTypeName) {
+  static const QString base = QStringLiteral("https://bigladdersoftware.com/epx/docs/25-1/input-output-reference/");
```
macumber: Let's define this base url in one spot, ideally in FindOpenStudioSDK.cmake

src/model_editor/InspectorGadget.cpp
```diff
@@ -467,9 +468,21 @@ void InspectorGadget::layoutText(QVBoxLayout* layout, QWidget* parent, openstudi
   if (level == AccessPolicy::LOCKED) {
     // string stripped(val);
     // std::replace(stripped.begin(), stripped.end(), '_', ' ');  // replace all '_' to ' '
-    auto* label = new QLabel(QString(val.c_str()), parent);
```
macumber, line 494: Shouldn't this also apply in the else block where `level != AccessPolicy::LOCKED`

macumber:
Looks pretty good, the one thing I would ask for is a tooltip which shows you the url before you click

src/model_editor/InspectorGadget.cpp
macumber: Hmm, this seems to apply to IGHeader only, I'm not sure why a header would ever be not locked.  Not an issue with this PR, just confusing existing code.

macumber: `index == -1` seems to be used for all headers, I'd suggest only doing the iddObjectDocUrl if `index == -1` to save some time, and add a comment that `// index == -1 indicates that this is a header row`

macumber: Actually this whole overload is only used for the headers, index could be removed as an input parameter.  I can do some clean up on this

macumber: Here is a cleaned-up function, if you want to replace it in your PR

```
void InspectorGadget::layoutHeaderText(QVBoxLayout* layout, QWidget* parent, const std::string& val, const std::string& comment) {
  auto* frame = new QFrame(parent);
  frame->setContentsMargins(0, 0, 0, 0);
  auto* hbox = new QHBoxLayout();
  frame->setLayout(hbox);
  frame->setObjectName("IGHeader");
  hbox->setSpacing(0);
  hbox->setContentsMargins(0, 0, 0, 0);

  // string stripped(val);
  // std::replace(stripped.begin(), stripped.end(), '_', ' ');  // replace all '_' to ' '
  auto* label = new QLabel(parent);
  label->setObjectName("IGHeader");
  label->setStyleSheet("font: bold");

  const QString typeName = QString::fromStdString(val);
  const QString docUrl = iddObjectDocUrl(typeName);
  if (!docUrl.isEmpty()) {
    label->setTextFormat(Qt::RichText);
    label->setOpenExternalLinks(true);
    label->setText(
      QStringLiteral(R"(<a href="%1" style="color: #ddeeff; font-weight: bold;">%2</a>)").arg(docUrl, typeName.toHtmlEscaped()));
  } else {
    label->setText(typeName);
  }

  // Qt::Alignment a = Qt::AlignHCenter;
  hbox->addWidget(label);

  hbox->addStretch();

  auto* commentText = new QLineEdit(QString(comment.c_str()), parent);
  commentText->setProperty(s_indexSlotName, -1);
  connect(commentText, &QLineEdit::textEdited, this, &InspectorGadget::IGcommentChanged);
  if (!m_showComments) {
    commentText->hide();
  }
  hbox->addWidget(commentText);
  commentText->setObjectName("IDFHeaderComment");

  layout->addWidget(frame);
}
```

src/model_editor/IddObjectDocUrl.hpp
macumber: Sorry I wasn't more explicit, follow the pattern in https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/utilities/OpenStudioApplicationPathHelpers.cxx.in

Add a method https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/utilities/OpenStudioApplicationPathHelpers.hpp and the value returned gets configured by a CMake variable into the source.

src/model_editor/CMakeLists.txt
```diff
@@ -103,6 +103,9 @@ else()
   target_compile_definitions(${target_name} PUBLIC model_editor_EXPORTS)
 endif()
 
+target_compile_definitions(${target_name} PRIVATE
```
macumber: Set BIGLADDERSOFTWARE_DOC_BASE_URL as a variable in https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/FindOpenStudioSDK.cmake, we have to update that file for each release so this will be visible when we do that.

Then add an accessor method to get the configured value in https://github.com/openstudiocoalition/OpenStudioApplication/blob/develop/src/utilities/OpenStudioApplicationPathHelpers.cxx.in

Ski90Moo:
Thanks for the feedback! Tooltips showing the full URL have been added to all clickable doc links — both in the inspector header (via `layoutHeaderText`) and in the `CollapsibleInspectorHeader`. Additionally, this branch has been extended to add doc links to several other parts of the UI: the left-sidebar category headers on the Loads, Constructions, Materials, and Other Schedules tabs, as well as section headers on the Site tab (Weather File, Design Days, Daylight Savings Time, Ground Temperatures), Life Cycle Costs tab, and the Facility tab (North Axis).

Ski90Moo:
Addressed the `BIGLADDERSOFTWARE_DOC_BASE_URL` feedback: removed the `target_compile_definitions` approach and added a `bigladdersoftwareDocBaseUrl()` accessor to `OpenStudioApplicationPathHelpers` following the existing `.cxx.in` pattern. The URL is configured from `BIGLADDERSOFTWARE_DOC_BASE_URL` in `FindOpenStudioSDK.cmake`, so it will be visible alongside the EnergyPlus version bump each release. All call sites updated to use `QString::fromStdString(openstudio::bigladdersoftwareDocBaseUrl())`.

src/openstudio_lib/BuildingInspectorView.cpp
```diff
@@ -437,8 +439,14 @@ BuildingInspectorView::BuildingInspectorView(bool isIP, bool displayAdditionalPr
   vLayout = new QVBoxLayout();
 
   label = new QLabel();
-  label->setText("North Axis: ");
-  label->setStyleSheet("QLabel { font: bold; }");
+  label->setTextFormat(Qt::RichText);
+  label->setOpenExternalLinks(true);
+  {
+    static const QString url =
+      QString::fromStdString(openstudio::bigladdersoftwareDocBaseUrl()) + "group-simulation-parameters.html#building";
```
macumber:
```suggestion
      QString::fromStdString(openstudio::bigladdersoftwareDocBaseUrl()) + iddObjectDocUrl("OS:Building");
```
It would be better to handle all of these using iddObjectDocUrl so all the url fragments are in one place.

src/model_editor/IddObjectDocUrl.hpp
```diff
@@ -0,0 +1,469 @@
...
+    {"OS:Table:IndependentVariableList", "group-performance-tables.html#tableindependentvariablelist"},
+    {"OS:Table:Lookup",                  "group-performance-tables.html#tablelookup"},
+  };
+  // clang-format on
+
+  auto it = urlMap.constFind(iddTypeName);
+  if (it != urlMap.constEnd()) {
+    return base + it.value();
+  }
+  return QString();
```
macumber: Let's log a warning here if the value is not found like `"Cannot find doc url for: " + iddTypeName`

src/model_editor/IddObjectDocUrl.hpp
```diff
@@ -0,0 +1,469 @@
...
+#include <QHash>
+
+#include "../utilities/OpenStudioApplicationPathHelpers.hpp"
+
+// Returns the BigLadder EnergyPlus I/O Reference URL for the given OS IDD type name
+// (e.g. "OS:ThermalZone"), or an empty string if none is known.
+// Base URL version is controlled by ENERGYPLUS_VERSION_MAJOR/MINOR in FindOpenStudioSDK.cmake.
+inline QString iddObjectDocUrl(const QString& iddTypeName) {
+  static const QString base = QString::fromStdString(openstudio::bigladdersoftwareDocBaseUrl());
+
```
macumber, line 23: IddFile does have a `std::vector<std::string> groups()` method, can we add a section so we can look up groups like "OpenStudio Simulation" -> "group-simulation-parameters.html"?

https://raw.githubusercontent.com/NatLabRockies/OpenStudio/refs/heads/develop/resources/model/OpenStudio.idd

src/model_editor/IddObjectDocUrl.hpp
macumber: Perfect, thanks!

src/model_editor/CMakeLists.txt
macumber: Fixed, thanks

macumber:
@Ski90Moo one thing I found is that you have to use the same version of clang-format that the CI is using, 18.1.3, other versions of clang-format don't find the same issues as CI.

Ski90Moo:
Thanks for the heads-up. The `clang-format` on this machine was the pip-installed 18.1.8 package shadowing the system LLVM install. Switched to 18.1.3 to match CI — both versions produce identical output on the PR-touched files, so the formatting commit pushed in `8a7f8d1` should satisfy the check once the workflows are approved to run.

Ski90Moo:
Addressed the remaining review comments:

**`layoutHeaderText` cleanup** — Adopted your refactored version: removed the `index` parameter, simplified the structure.

**`BuildingInspectorView.cpp`** — Switched to `iddObjectDocUrl(QStringLiteral("OS:Building"))` so the URL fragment lives in one place alongside all the others. Swapped the `OpenStudioApplicationPathHelpers` include for `IddObjectDocUrl.hpp`.

**Warning log** — Added `qWarning() << "Cannot find doc url for: " + iddTypeName;` before the empty return in `iddObjectDocUrl`.

**Group URL lookup** — Added `iddGroupDocUrl(groupName)` to `IddObjectDocUrl.hpp` mapping OpenStudio.idd `\group` names (e.g. `"OpenStudio Simulation"`, `"OpenStudio Materials"`) to their BigLadder page URLs. Groups that span multiple EnergyPlus chapters (e.g. `"OpenStudio HVAC"`) are intentionally omitted. All 395 URL references pass `check_doc_urls.py`.

src/model_editor/IddObjectDocUrl.hpp
```diff
@@ -0,0 +1,503 @@
...
+    {"Energy Management System (EMS)",               "group-energy-management-system-ems.html"},
+    {"User Defined HVAC and Plant Component Models", "group-user-defined-hvac-and-plant-component.html"},
+  };
+  // clang-format on
+
+  auto it = groupMap.constFind(groupName);
+  if (it != groupMap.constEnd()) {
+    return base + it.value();
+  }
+  return {};
```
macumber, line 527:
```suggestion
  qWarning() << "Cannot find doc url for: " + groupName;
  return {};
```

src/openstudio_lib/ConstructionsView.cpp
```diff
@@ -12,35 +12,39 @@
...
-  //result.push_back(std::make_pair<IddObjectType, std::string>(IddObjectType::OS_Construction_WindowDataFile, "Window Data File Constructions"));
-
-  return result;
+std::vector<std::tuple<IddObjectType, std::string, QString>> ConstructionsView::modelObjectTypesNamesAndUrls() {
+  static const QString base = QString::fromStdString(openstudio::bigladdersoftwareDocBaseUrl());
+  static const QString sce = base + QStringLiteral("group-surface-construction-elements.html");
+
+  using T = std::tuple<IddObjectType, std::string, QString>;
+  return {
+    T{IddObjectType::OS_Construction, "Constructions", sce + "#construction-000"},
```
macumber:
```suggestion
    T{IddObjectType::OS_Construction, "Constructions", iddObjectDocUrl(QStringLiteral("OS:Construction"),
```
These urls are already defined in IddObjectDocUrl.hpp, we should use those definitions rather than duplicate them here.  Same thing for all the other urls in this file and others such as:

- https://github.com/openstudiocoalition/OpenStudioApplication/pull/876/changes#diff-041047d3606ed5e7fec54dcfe4e8e7ce2f5720433717aaa1e00936250e7140b6R105
- https://github.com/openstudiocoalition/OpenStudioApplication/pull/876/changes#diff-e35d718a8caea938a126dcce1d7f7baae6e2878d2325aa60c129adc74f6254e0R159
- https://github.com/openstudiocoalition/OpenStudioApplication/pull/876/changes#diff-25e93e6585f3692c885d623d1223853ca5b3362b49d23fb99e8bbeb6d1ba357dR53
- etc

src/openstudio_lib/BuildingInspectorView.cpp
Ski90Moo: Addressed in 2b581104 — the inline URL fragments were consolidated into `iddObjectDocUrl()`, so the call site now just uses `iddObjectDocUrl(QStringLiteral("OS:Building"))` directly.

src/openstudio_lib/ConstructionsView.cpp
Ski90Moo: Thank you macumber, agreed. Done in 42c02c26 — added 18 missing type→URL entries to `IddObjectDocUrl.hpp` (Construction variants, WindowMaterial variants, `Schedule:File`, `GroundTemperature:Shallow/Deep`, `RunPeriodControl:DaylightSavingTime`, `ConvergenceLimits`, `ZoneCapacitanceMultiplier:ResearchSpecial`, Output types, LifeCycleCost types) and replaced all `bigladdersoftwareDocBaseUrl() + hardcoded fragment` patterns across 9 view files (`ConstructionsView`, `MaterialsView`, `LoadsView`, `ScheduleOthersView`, `SimSettingsView`, `GroundTemperatureView`, `LifeCycleCostsTabView`, `LocationTabView`, `YearSettingsWidget`) with `iddObjectDocUrl()` / `iddGroupDocUrl()` calls.

macumber:
@jmarrec do you have thoughts on this PR?  Both this PR and the BCL one rely on an external webpage which we don't have control over.  I have thought that it might be better to make the URLs configurable in preferences rather than hard coded.  However, if the URLs ever change it is also likely that the API/site layout would change which would break the integration.  I am leaning towards something like 1) Make the URLs configurable, 2) Test the URLs at startup and disable functionality if they are not available.   What do you both think?  That could be implemented on a separate PR if we want to do it.  I think this PR is setup for it well, if we return an empty URL when the integration test fails the tooltips just won't be added.

jmarrec:
I like the idea of adding some help! 

But I think I'm quite against linking to the Big Ladder website because like you said we have zero control over it. The URL may change, the links could be broken in a new version, etc. The BCL endpoint changing recently is a pain that makes it quite clear this is something we should avoid.

I've already build the E+ HTML documentation recently, it's quite simple to do with pandoc really. I'm not quite sure if we are allowed to host that on a domain we control or not though (license wise), but this is something we should perhaps investigate.

macumber:
Maybe we just build the html and then ship with the app?

Ski90Moo:
Thanks for digging into this, @macumber and @jmarrec — these are fair concerns and I want to share the motivation behind the feature so the tradeoff is clearer.

The original driver was actually the in-progress i18n work (#873): once the GUI is translated, a non-English-speaking user is still stuck with English-only EnergyPlus/OpenStudio documentation. Rather than have OpenStudio take on the burden of translating that documentation ourselves, the idea is to get the user *to* the relevant doc page quickly so they can use their own browser's built-in translator (Chrome, Edge, etc. all do this well now) — no extra maintenance burden on us, and the user gets a translation tuned to their specific language and reading level.

On the "we don't control BigLadder" concern — that's true, but I'd note a few things that make this a different kind of dependency than the BCL endpoint:
- These links point to *archived, version-specific* EnergyPlus documentation pages (e.g., a specific E+ version's input/output reference). They aren't a live API tied to current data — once a version's docs are published, they don't get rewritten or recategorized.
- BigLadder has hosted these archived docs at stable URLs for years without restructuring them, even as new EnergyPlus versions ship (the new version's docs just get a new path; old paths keep working).
- The PR is already structured so a missing/changed mapping degrades gracefully — `iddObjectDocUrl()`/`iddGroupDocUrl()` return an empty string when there's no entry, so no tooltip/link is shown rather than a broken one.

I think @macumber's suggestion (configurable base URL + an availability check at startup that disables the feature gracefully) is a great safety net regardless of which docs we ultimately link to, and could be layered on without much rework — happy to tackle that as a fast-follow on this PR or a separate one.

@jmarrec's idea of bundling locally-built E+ HTML docs is appealing for full independence, but it's a much larger scope change (build pipeline, packaging size, licensing question you raised, and keeping it in sync with the bundled E+ version). If that's the direction the project wants to go long-term, I'm happy to help, but I'd suggest it's worth treating as its own initiative rather than blocking this PR — which, as currently designed, fails safe if the links ever do break.

Let me know how you'd like to proceed — happy to adjust scope based on your call.

jmarrec:
I am going to do some legwork to inquire whether we can host it somewhere (we've discussed with the E+ dev team to have it on energyplus.net or readthedocs in the past, and otherwise I'll see if we can do it on a domain we control here) . I'll keep you posted.

Ski90Moo:
> I am going to do some legwork to inquire whether we can host it somewhere (we've discussed with the E+ dev team to have it on energyplus.net or readthedocs in the past, and otherwise I'll see if we can do it on a domain we control here) . I'll keep you posted.

That would be great!  Thank you @jmarrec.

macumber:
@Ski90Moo let's include this in the RC2

In order to avoid depending on more online sources (which have caused us problems), let's update this to work with installed docs in the package.  To do this, I think the basic steps are:
- Add a FindEnergyPlus CMake that downloads and finds EnergyPlus
    - Try to make it where all we have to do is update the EnergyPlus version 
    - https://github.com/NatLabRockies/OpenStudio/blob/develop/CMakeLists.txt#L668
    - https://github.com/NatLabRockies/OpenStudio/blob/develop/CMake/FindEnergyPlus.cmake
- Build the EnergyPlus Input Output documentation to html
    -  https://github.com/NatLabRockies/EnergyPlus/tree/develop/doc/input-output-reference
    - https://github.com/NatLabRockies/EnergyPlus/blob/develop/.github/workflows/build_documentation.yml
    - https://github.com/NatLabRockies/EnergyPlus/blob/develop/CMakeLists.txt
    - https://github.com/NatLabRockies/EnergyPlus/blob/develop/doc/readthedocs/README.md
- Add install code to install the html docs in our package under EnergyPlus/doc
- Update your code to point to local base URL derived from the paths in OpenStudioApplicationPathHelpers.hpp

@jmarrec might be able to give you a little more specifics

Ski90Moo:
@jmarrec — thanks again for the `ep_doc_html` export, it's been really useful. We've gone ahead and wired it in as a stopgap per @macumber's suggestion: the app now downloads your export, from the link you provided, at configure time (MD5-verified), installs it alongside the app, and all of `IddObjectDocUrl.hpp`'s doc links now point at the local copy instead of bigladdersoftware.com. ([commit](https://github.com/Ski90Moo/OpenStudioApplication/commit/18d3310f) on our fork's `feat/doc-links-inspector`)

Confirmed working: real cross-page hyperlinks, native MathML equations (no JS/bitmaps), and all 393 of our existing anchor references resolve correctly against your export's HTML.

A few questions as we look at making this permanent rather than a one-off download:

1. Can we run the build pipeline behind this ourselves, in CI or at release time, rather than depending on a manual export each time?
2. Is the pipeline tied to a specific EnergyPlus tag/version the way the official releases are, so we could pin our build to whatever version the OpenStudio SDK bundles?
3. The output currently pulls Bootstrap/Bootstrap-Icons/Highlight.js from jsdelivr/cdnjs CDNs — we'd want to vendor those locally so the docs work fully offline (same reason we're moving off bigladdersoftware.com). Any objection to that, or a reason it's CDN-hosted today that we should know about?
4. Found one broken internal link in the export: `1.3-input-output-descriptions-document.html` (and 4 other pages) link to `1.4-sec:api-usage.html` (colon), but the actual file is `1.4-sec_api-usage.html` (underscore) — looks like a label-sanitization mismatch between the cross-reference generator and the output-filename generator in the Pandoc pipeline. Confirmed it's the only broken local link out of 8,519 internal cross-references in the export.

jmarrec:
> * Can we run the build pipeline behind this ourselves, in CI or at release time, rather than depending on a manual export each time?
> * Is the pipeline tied to a specific EnergyPlus tag/version the way the official releases are, so we could pin our build to whatever version the OpenStudio SDK bundles?

Yes, you do want to build documentation for a specific E+ tag.
Don't worry about doing that yourself, I need to figure out how I upstream that. And there are 2 releases of EnergyPlus per year, so it's fine to do manually if needed (I don't have to, I wired it up into github actions already)


> * The output currently pulls Bootstrap/Bootstrap-Icons/Highlight.js from jsdelivr/cdnjs CDNs — we'd want to vendor those locally so the docs work fully offline (same reason we're moving off bigladdersoftware.com). Any objection to that, or a reason it's CDN-hosted today that we should know about?

The HTML docs were meant to be online, so CDN made sense. I also happen to have a single-page version where it was nice to have everything in one file (images can be base64 encoded)

Anyways, that's easy to change, download a few files, do a global sed.

> * Found one broken internal link in the export: `1.3-input-output-descriptions-document.html` (and 4 other pages) link to `1.4-sec:api-usage.html` (colon), but the actual file is `1.4-sec_api-usage.html` (underscore) — looks like a label-sanitization mismatch between the cross-reference generator and the output-filename generator in the Pandoc pipeline. Confirmed it's the only broken local link out of 8,519 internal cross-references in the export.

Yeah I actually need to figure that out, I hit this error when uploading to github artifact too (the colon was rejected for upload as it's not NTFS compliant)

PR #883, opened 2026-06-16
Add Apply to Selected to Facility Shading tab Construction and Schedule columns

### LINKED ISSUE #882: Add Apply to Selected to the Facility Shading tab
Enhancement Request
--------------

Add Apply to Selected buttons for the columns on this tab so that the checkboxes can also be used to apply across multiple selected objects.

Ski90Moo:
## Summary
- Enables the "Apply to Selected" button on the Construction Name and Transmittance Schedule Name columns of the Facility Shading tab
- The `showColumnButton` flag was `false` for both drop zone columns; changed to `true` to match the pattern used by other tabs

Closes #882

## Test plan
- [ ] Open a model with multiple shading surface groups
- [ ] Navigate to Facility tab → Shading subtab
- [ ] Select multiple rows using the checkboxes
- [ ] Click a cell in the Construction Name or Transmittance Schedule Name column
- [ ] Verify the "Apply to Selected" button appears and applies the value to all selected rows


macumber:
Let's wait on this for RC2

PR #891, opened 2026-07-15
Add hot-gas reheat unitary system (Coil:Heating:Desuperheater)

Ski90Moo:
## Summary
- Adds "Unitary - Single Speed DX cooling - Elec heat - CAV - Desuperheater reheat" to the default HVAC library: the existing single-speed DX cooling coil paired with a new `Coil:Heating:Desuperheater` as the supplemental/reheat coil under `CoolReheat` dehumidification control, so reheat is partially sourced from the cooling coil's waste heat instead of just resistance heat.
- Fixes `AirLoopHVACUnitarySystem`/loop cloning (both "drag from library" and "Copy System"): `ModelObject::clone()` only remaps true parent-child relationships. Lateral (sibling-to-sibling) object-list references — like `CoilHeatingDesuperheater::heatingSource()`, or a `SetpointManager:MixedAir`'s node fields — are left broken on the clone, sometimes stale, sometimes cleared outright. Added a generic post-clone reference fixup (`buildCloneHandleMap` + `remapLateralReferences` in `HVACSystemsController.cpp`, using `IddObject::objectLists()`/`WorkspaceObject::getTarget()`/`setPointer()`) rather than a Desuperheater-specific special case.
- The traversal walks `children()`, a loop's `supplyComponents()`/`demandComponents()`, and an outdoor air system's `oaComponents()`/`reliefComponents()` as independently size-matched groups, since branch equipment (units on a loop, or equipment like an evaporative cooler on an OA system's branches) isn't reachable via `children()` alone, and demand-side counts can legitimately differ between original and clone since connected thermal zones are never duplicated.

## Context
Modeling dehumidification via hot-gas/desuperheater reheat is a recurring real-world need — see [this UnmetHours thread](https://unmethours.com/question/103316/how-to-make-coilcoolingdxtwostagewithhumiditycontrolmode-work/), where `CoilCoolingDXTwoStageWithHumidityControlMode` proved awkward to control directly and commenters pointed toward `Coil:Heating:Desuperheater` as a workaround. Worth noting the thread's caveat: a desuperheater coil reclaims waste heat from the refrigerant's superheat and doesn't affect compressor performance, so it's a lower-fidelity approximation of "true" hot-gas reheat, not an exact substitute — but it's a standard, supported EnergyPlus object and a reasonable default-library addition for this use case.

## Proof this was broken before this fix
You don't need the new desuperheater system to see the underlying bug — it reproduces with a completely stock model:
1. In the app, `File > Examples > Example Model` to generate a fresh example model.
2. Go to the HVAC Systems tab, select the one airloop, click "Copy System."
3. Select the copy and open its outdoor air system. It has an evaporative cooler with a `SetpointManager:MixedAir` attached — but on the *copy*, that setpoint manager's Reference Setpoint / Fan Inlet / Fan Outlet Node fields are all empty (shown as warning icons in the inspector), even though the original's are filled in correctly.
4. Run a simulation on the copy: EnergyPlus fails immediately with a fatal error (missing required properties on that object) before any calculation happens.

That's `ModelObject::clone()`'s lateral-reference bug on a totally unmodified example model — this PR's fixup resolves it for both this stock case and the new desuperheater system.

## Test plan
- [x] `cmake --build . --target openstudio_lib --config Release` succeeds
- [x] Dragged the new system onto a zone in the app and ran the simulation successfully
- [x] Used "Copy System" on that airloop, reassigned zones on the copy, and ran the simulation successfully
- [x] Verified against the SDK directly (Ruby) for same-model/cross-model clone scenarios, and against `OpenStudio::Model.exampleModel`'s stock outdoor-air-system evaporative cooler + `SetpointManager:MixedAir`
- [x] Confirmed live in the rebuilt app that "Copy System" correctly re-wires node references on that stock example airloop too

### DIFF
--- a/src/openstudio_lib/HVACSystemsController.cpp
+++ b/src/openstudio_lib/HVACSystemsController.cpp
@@ -626,7 +626,7 @@ void remapLateralReferences(const model::ModelObject& original, model::ModelObje
 
 // `original` is the subtree that was cloned; `clonedObject` is its clone. Re-links any lateral
 // reference in the clone that still points into (or should point into) the original subtree.
-void fixupClonedReferences(const model::ModelObject& original, model::ModelObject clonedObject) {
+void fixupClonedReferences(const model::ModelObject& original, const model::ModelObject& clonedObject) {
   std::map<Handle, model::ModelObject> handleMap;
   buildCloneHandleMap(original, clonedObject, handleMap);
   remapLateralReferences(original, clonedObject, handleMap);
--- a/src/openstudio_lib/HVACSystemsController.cpp
+++ b/src/openstudio_lib/HVACSystemsController.cpp
@@ -92,10 +92,6 @@
 #include <openstudio/model/CoilHeatingElectric_Impl.hpp>
 #include <openstudio/model/CoilHeatingWater.hpp>
 #include <openstudio/model/CoilHeatingWater_Impl.hpp>
-#include <openstudio/model/CoilHeatingDesuperheater.hpp>
-#include <openstudio/model/CoilHeatingDesuperheater_Impl.hpp>
-#include <openstudio/model/AirLoopHVACUnitarySystem.hpp>
-#include <openstudio/model/AirLoopHVACUnitarySystem_Impl.hpp>
 #include <openstudio/model/ParentObject.hpp>
 #include <openstudio/model/ParentObject_Impl.hpp>
 #include <openstudio/model/Model.hpp>
@@ -131,6 +127,8 @@
 #include <openstudio/utilities/core/Assert.hpp>
 #include <openstudio/utilities/core/Compare.hpp>
 #include <openstudio/utilities/idd/OS_ComponentData_FieldEnums.hxx>
+
+#include <map>
 
 #include <QMessageBox>
 #include <QTimer>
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -39,4 +52,61 @@ clang_format.patch
 conan-cache
 .ccache
 CMakeUserPresets.json
-__pycache__
\ No newline at end of file
+
+# This repo is built in-place at the repo root (not an out-of-source build/ dir), so
```
macumber: I suggest not building in-source

.gitignore
```diff
@@ -25,9 +25,22 @@ relwithdebinfo/
 profile/
 /super-build-shared/
 /super-build-static/
+
 *.sublime-workspace
+
+# Qt moc/rcc-generated files, wherever CMake drops them in the tree
```
macumber: This is also happening because you are building in-source.

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -519,6 +523,117 @@ void HVACLayoutController::addLibraryObjectToTopLevel(const OSItemId& itemId) {
...
+  // ParentObject::children() is true ownership (e.g. a UnitarySystem's fan/coils) but does *not*
+  // include HVACComponents placed on a Loop's supply/demand branches -- a UnitarySystem sitting on
+  // an AirLoopHVAC's supply branch is reachable via supplyComponents(), not children(). The same
+  // gap exists one level deeper: equipment on an AirLoopHVACOutdoorAirSystem's outdoor-air/relief
+  // branches (e.g. an evaporative cooler, or a SetpointManager:MixedAir sitting on one of those
+  // nodes) is reachable only via oaComponents()/reliefComponents(), not children() either. Cloning
+  // a whole loop (the "copy system" toolbar action) needs all of these, or the branch equipment --
+  // and anything lateral-referenced from inside it -- is never visited at all.
+  //
+  // Returned as separate groups, each matched and recursed into independently: cloning a loop does
```
macumber: This makes sense to me but I wonder two things:

1. Why has this not been an issue before?  Does the current workflow just make you have to go through the new airloop and reattach things to each other?
2. Should this be proposed to the SDK as an improvement to AirLoopHVAC::clone?  Or maybe even ModelObject::clone?

.gitignore
Ski90Moo: Well that is embarrassing.  Thank you for the tip @macumber

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Good questions.

1. It actually has been latent — `clone()` has never remapped lateral (sibling-to-sibling) object-list references, only true parent-child ownership edges. Both the desuperheater case and this OA-branch case fail loudly in E+ when hit. The difference is exposure: `heatingSource()` is core to every use of the new desuperheater template, so it's unmissable. The OA-branch gap (equipment on an `AirLoopHVACOutdoorAirSystem`'s oa/relief branch, reachable only via `oaComponents()`/`reliefComponents()`) only fires for a narrower setup — an airloop whose OA system has its own branch equipment with a `SetpointManager:MixedAir` targeting those nodes, like the stock File > Examples > Example Model. Anyone who hit that fatal error after "Copy System" most likely just manually re-filled the blank node fields — indistinguishable from the zone-reassignment step users already expect to do after copying — rather than recognizing it as an app bug.
2. Agreed that fixing this inside `ModelObject::clone()`/`AirLoopHVAC::clone()` itself would be the more correct place — it'd benefit every SDK client, not just this app. But doing it "correctly" there is a bigger change: clone() currently clones children one at a time via `setParent()` rather than batch-cloning the whole subtree into one shared handle map first, so fixing it at that level means restructuring clone's own orchestration for every model object type, not just HVAC. That's more design/test surface than fits in this PR. I'd like to land this app-level fix now (it's scoped and tested, and it also fixes the example-model bug above), and open a tracking issue on the SDK repo proposing the real fix — I can file that and link it here if that sounds right.

.gitignore
Ski90Moo: Good catch — that commit only existed because I was building in-source, which BUILDING.md doesn't actually recommend. Switched to an out-of-source build and reverted it (Ski90Moo#4), then rebased this branch onto develop to drop it from this PR entirely. Diff is now just the feature.

.gitignore
Ski90Moo: Same as above — resolved by switching to an out-of-source build; this whole commit is dropped from the PR now.

src/openstudio_lib/HVACSystemsController.cpp
macumber: Sounds good, please do open an SDK issue to track this.  If the SDK fixes the issue we can remove this.  But this fix will be useful for now.

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -527,7 +641,9 @@ void HVACLayoutController::addLibraryObjectToModelNode(const OSItemId& itemId, m
   object = doc->getModelObject(itemId);
   if (object) {
     if (!doc->fromModel(itemId)) {
+      model::ModelObject original = object.get();
       object = object->clone(comp.model());
+      fixupClonedReferences(original, object.get());
```
macumber, line 537: I would suggest moving `fixupClonedReferences` to src/utilities and adding some unit tests

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -519,6 +523,116 @@ void HVACLayoutController::addLibraryObjectToTopLevel(const OSItemId& itemId) {
...
+  }
+}
+
+// For every lateral object-list field on `original` whose target was itself cloned (i.e. has an
+// entry in handleMap), forces the corresponding field on `clonedObject` to point at that clone.
+// Fields are read from `original`, not from `clonedObject`: clone() doesn't necessarily leave a
+// lateral reference pointing at the stale original object -- for CoilHeatingDesuperheater's
+// heatingSource() it clears the field outright -- so the only reliable source of "what this
+// field is supposed to point at" is the original.
+void remapLateralReferences(const model::ModelObject& original, model::ModelObject clonedObject,
```
macumber: clonedObject can be passed by reference

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -519,6 +523,116 @@ void HVACLayoutController::addLibraryObjectToTopLevel(const OSItemId& itemId) {
...
+        remapLateralReferences(originalChildren[i], cloneChildren[i], handleMap);
+      }
+    }
+  }
+}
+
+// `original` is the subtree that was cloned; `clonedObject` is its clone. Re-links any lateral
+// reference in the clone that still points into (or should point into) the original subtree.
+void fixupClonedReferences(const model::ModelObject& original, const model::ModelObject& clonedObject) {
+  std::map<Handle, model::ModelObject> handleMap;
```
macumber: Can use std::unordered_map since ordering isn't important

src/openstudio_lib/HVACSystemsController.cpp
```diff
@@ -519,6 +523,116 @@ void HVACLayoutController::addLibraryObjectToTopLevel(const OSItemId& itemId) {
...
+  return groups;
+}
+
+// Walks `original` and `clone` in parallel -- childSubtreeObjectGroups() order is deterministic
+// and preserved by clone() -- building a map from each original object's handle to its
+// corresponding clone.
+void buildCloneHandleMap(const model::ModelObject& original, const model::ModelObject& clone, std::map<Handle, model::ModelObject>& handleMap) {
+  handleMap.emplace(original.handle(), clone);
+
+  std::vector<std::vector<model::ModelObject>> originalGroups = childSubtreeObjectGroups(original);
```
macumber: Can you move this recursion into childSubtreeObjectGroups so you just call childSubtreeObjectGroups once and it gets all the subtrees in one call?

src/openstudio_lib/HVACSystemsController.cpp
macumber: Just add a few comments and suggestions to improve efficiency since there is a lot of looping.  I'd move this to utilities since it is a candidate SDK method.

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Done in 1256782 — `remapLateralReferences` now takes `model::ModelObject&` instead of by value.

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Done in 1256782 — switched to `std::unordered_map<Handle, model::ModelObject, boost::hash<boost::uuids::uuid>>` (matching the pattern already used for `Handle` keys in `Workspace_Impl.hpp`, since `Handle` has no `std::hash` specialization).

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Done in 1256782 — merged the duplicated tree-walk from `buildCloneHandleMap`/`remapLateralReferences` into a single recursive pass (`collectClonedObjectPairs`) that's called once and returns the flat list of matched original/clone pairs; both the handle-map build and the reference remap now just iterate that list.

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Done in 1256782 — moved `fixupClonedReferences` (and its now-private helpers) to `src/utilities/CloneFixup.{hpp,cpp}`, and added `src/utilities/test/CloneFixup_GTest.cpp` with a test covering the `CoilHeatingDesuperheater::heatingSource()` case plus a no-op sanity check.

src/openstudio_lib/HVACSystemsController.cpp
Ski90Moo: Filed NREL/OpenStudio#5636 to track the underlying clone() gap upstream.

